PR #5, opened 2018-09-07
Feature/send

mat-hek:


lib/unifex/native_code_generator.ex
```diff
@@ -47,6 +49,35 @@ defmodule Unifex.NativeCodeGenerator do
   * `I` indents every line of string
   """
   @spec sigil_g(String.t(), charlist()) :: String.t()
+  def sigil_g(content, 'j(' ++ flags) when is_list(content) do
```
bblaszkow06: Add description of new features to docs

lib/unifex/native_code_generator.ex
```diff
@@ -47,6 +49,35 @@ defmodule Unifex.NativeCodeGenerator do
...
+    {joiner, ')' ++ flags} = flags |> Enum.split_while(&([&1] != ')'))
+    content = content |> Enum.join("#{joiner}")
+    sigil_g(content, flags)
+  end
+
+  def sigil_g(content, 'n' ++ flags) when is_list(content) do
```
bblaszkow06: BTW, why on earth would you apply sigils on list? Not only it contradicts the specs, but also is super weird. Why don't you extract the functionality for lists into separate function?

lib/unifex/native_code_generator.ex
```diff
@@ -99,8 +87,33 @@ defmodule Unifex.NativeCodeGenerator do
...
+
+  It supports all the flags supported by `sigil_g/2` and the following ones:
+  * `j(joiner)` - joins list of strings using `joiner`
+  * n - alias for `j(\\n)`
+
+  If passed a list and flags supported by `sigil_g/2`, each flag will be executed
+  on each element of the list, until the list is joined by using `j` or `n` flag.
+  """
+  @spec g(String.Chars.t() | [String.Chars.t()], charlist()) :: String.t() | [String.t()]
+  def g(content, 'j(' ++ flags) when is_list(content) do
```
bblaszkow06: Please, rename it to something meaningful like `formatter`. Even `gen` would be better.

PR #8, opened 2018-10-08
Fixes required by ffmpeg h264

bblaszkow06:


lib/unifex/base_type/unsigned.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Unifex.BaseType.Unsigned do
+  @moduledoc """
+  Module implementing `Unifex.BaseType` behaviour for Unifex state.
```
mat-hek: not really...

PR #9, opened 2018-10-09
added ability to mark functions as dirty

mat-hek:


lib/unifex/specs.ex
```diff
@@ -0,0 +1,153 @@
...
+                                         long bytes_used, long sample_rate, int channels);
+      UNIFEX_TERM decode_frame_result_error_buflen(UnifexEnv* env);
+      UNIFEX_TERM decode_frame_result_error_malformed(UnifexEnv* env);
+      UNIFEX_TERM decode_frame_result_error_recoverable(UnifexEnv* env, int bytes_to_skip);
+
+  See docs for appropriate macros for more details.
+  """
+
+  @doc """
+  Macro used for defining module that is to export native functions to Elixir world.
```
bblaszkow06: That sentence is quite hard to understand.
* `Macro used for defining module` -> `Defines module` sounds better and isn't that complicated.
* `that is to export native functions to Elixir world.` Again - overly complicated. I would rephrase it to something like `that exports native functions ..`

lib/unifex/specs.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  The module needs to be defined manually, but it can `use` `Unifex.Loader` to
+  have functions declared with `spec/1` automatically defined.
+  """
+  defmacro module(module) do
+    store_config(:module, module)
+  end
+
+  @doc """
+  Macro used for defining native function specification.
```
bblaszkow06: -> `Defines` or `Macro that defines` IMO sounds better

lib/unifex/specs.ex
```diff
@@ -0,0 +1,153 @@
...
+  when type is one of:
+  - `:cpu` - marks function as CPU-bound (maps to the `ERL_NIF_DIRTY_JOB_CPU_BOUND` erlang flag)
+  - `:io` - marks function as IO-bound (maps to the `ERL_NIF_DIRTY_JOB_IO_BOUND` erlang flag)
+  """
+  defmacro dirty(type, funs) when type in [:cpu, :io] and is_list(funs) do
+    store_config(:dirty, funs |> Enum.map(&{&1, type}))
+  end
+
+  @doc """
+  Macro used for defining what can be sent from the native code to elixir processes.
```
bblaszkow06: How about `Defines types of messages that can be ..`

PR #22, opened 2018-12-27
Allow to define callbacks for NIF lifecycle hooks

bblaszkow06:
I needed a way to provide load/unload callbacks as described [here](http://erlang.org/doc/man/erl_nif.html#initialization) and here's my proposal.

BTW, if we have some spare time `native_code_generator` could use some refactoring as it's getting quite big


lib/unifex/native_code_generator.ex
```diff
@@ -340,13 +357,46 @@ defmodule Unifex.NativeCodeGenerator do
...
       #{state_type} state = (#{state_type}) value;
       #{generate_unifex_env()}
       handle_destroy_state(unifex_env, state);
     }
+    """
+  end
 
-    static int load(ErlNifEnv* env, void** priv_data, ERL_NIF_TERM load_info) {
+  defp generate_nif_lifecycle_callbacks_declarations(callbacks) do
+    callbacks
```
mat-hek: support default callback names

lib/unifex/native_code_generator.ex
```diff
@@ -340,13 +357,46 @@ defmodule Unifex.NativeCodeGenerator do
...
+    end)
+  end
+
+  defp generate_nif_lifecycle_callbacks(nil, _callbacks) do
+    ~g""
+  end
+
+  defp generate_nif_lifecycle_callbacks(_module, callbacks) do
+    load_result =
+      case callbacks[:load] do
```
mat-hek: all these cases look too verbose and violate DRY

bblaszkow06: I disagree. Being verbose is not a flaw (and it could have been less verbose if not for a formatter) and the only common thing for the cases is changing `nil` to `""` and passing a name to a code template in 2 out 3 cases. Not worth the effort.

lib/unifex/specs.ex
```diff
@@ -139,7 +139,14 @@ defmodule Unifex.Specs do
     `void on_unload(UnifexEnv *env, void * priv_data)`
 
   """
-  defmacro callback(hook, fun) when hook in [:load, :upgrade, :unload] and is_atom(fun) do
+  defmacro callback(hook, fun \\ nil) when hook in [:load, :upgrade, :unload] and is_atom(fun) do
+    fun =
+      if fun == nil do
```
mat-hek: I think that creating a separate one-parameter definition invoking the current one would look better

PR #30, opened 2020-01-17
C++ support

FelonEkonom:


lib/unifex/interface_IO.ex
```diff
@@ -26,11 +26,20 @@ defmodule Unifex.InterfaceIO do
     File.mkdir_p!(out_dir_name)
     out_base_path = Path.join(out_dir_name, name)
     File.write!("#{out_base_path}.h", header)
-    File.write!("#{out_base_path}.c", source)
+
+    src_f_ext =
+      if File.exists?(Path.join(dir, name <> ".cpp")) do
```
mat-hek: we agreed to always store both .c and .cpp files

lib/unifex/native_code_generator.ex
```diff
@@ -164,13 +168,24 @@ defmodule Unifex.NativeCodeGenerator do
...
   defp generate_source(name, module, functions, results, dirty_funs, sends, callbacks) do
     ~g"""
     #include "#{name}.h"
 
+    ErlNifResourceType *UNIFEX_PAYLOAD_GUARD_RESOURCE_TYPE;
+
+    UnifexPayload *unifex_payload_alloc(
+            UnifexEnv *env, UnifexPayloadType type, unsigned int size) {
+      return unifex_payload_alloc_template(env, type, size, &UNIFEX_PAYLOAD_GUARD_RESOURCE_TYPE);
+    }
```
mat-hek: extract to a separate function

PR #33, opened 2020-02-11
Cnode support

FelonEkonom:
TODO:
- [x] Refactor cnode version of unifex.c
- [x] Refactor CodeGenerationMode
- [x] Refactor Specs
- [x] Support custom main functions
- [x] Add docs and specs
- [x] Support binary payloads


lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+  defstruct @enforce_keys
+
+  @type t :: %__MODULE__{
+          server: pid,
+          node: node
+        }
+
+  @type on_start_t :: {:ok, t} | {:error, :spawn_cnode | :connect_to_cnode}
+
+  def cast_on_start_t({:ok, %Bundlex.CNode{} = bundex_cnode}) do
```
mat-hek: I would keep Bundlex.CNode within Unifex.CNode instead

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+  def cast_cnode(%__MODULE__{server: server, node: node}) do
+    %Bundlex.CNode{
+      server: server,
+      node: node
+    }
+  end
+
+  @doc """
+  Calls Bundlex.CNode functions.
+  Look at Bundlex.CNode docs, to see more.
```
mat-hek: This says nothing :P btw, read https://hexdocs.pm/elixir/writing-documentation.html, especially 'recommendations' section

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+    content
+  end
+
+  @doc """
+      Sends to CNode message containing data about which function should be called and with which args,
+      then waits timeout miliseconds to receive and unpack function result
+  """
+  @spec remote_call(t, fun_name :: atom, args :: list, timeout :: non_neg_integer | :infinity) ::
+          response :: term
+  def remote_call(%__MODULE__{} = unifex_cnode, fun_name, args \\ [], timeout \\ 5000) do
```
mat-hek: would call it just `call`

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+    content
+  end
+
+  def unpack_message({:send, content}) do
+    content
+  end
+
+  @doc """
+      Sends to CNode message containing data about which function should be called and with which args,
+      then waits timeout miliseconds to receive and unpack function result
```
mat-hek: Put more focus on the point (what?) instead of the process (how?). The end-user doesn't need to know that some message is sent and the result is packed anyhow.

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  @doc """
+      Sends to CNode message containing data about which function should be called and with which args,
+      but don't wait on any message back.
+      
+      Should be used, when called CNode function sends messages (look at .spec.exs), instead of return value
+  """
+  @spec start_running(t, fun_name :: atom, args :: list) :: :ok
+  def start_running(%__MODULE__{} = unifex_cnode, fun_name, args \\ []) do
```
mat-hek: calling this `cast` would be more erlang/elixir way

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,168 @@
...
+
+  @doc """
+      Waits timeout miliseconds on message sent from specific UnifexCNode.
+  """
+  @spec receive_msg(t, timeout :: non_neg_integer | :infinity) ::
+          response :: term | {:error, :time_left}
+  def receive_msg(%__MODULE__{node: node}, timeout \\ 5000) do
+    receive do
+      {^node, {:send, content} = response} ->
+        response |> unpack_message
```
mat-hek: The sent messages should need no unpacking

lib/unifex/code_generator/cnode_code_generator.ex
```diff
@@ -138,23 +162,46 @@ defmodule Unifex.CodeGenerator.CNodeCodeGenerator do
...
+              |> Keyword.get_values(:label)
+              |> (fn
+                    labels when labels != [] ->
+                      labels
+
+                    _ ->
+                      [head | _tail] = specs |> Tuple.to_list()
+                      [head]
+                  end).())
+  end
```
mat-hek: This is extremely unreadable

FelonEkonom:
New functionalities introduced in Unifex so far:
 - generating boilerplate code for CNodes based on .spec.exs files (as in NIFs). Generated code includes implementation of send_* and [fun_name]_result_* functions
 - C++ support (in another PR) for NIFs and CNodes
 - support for using state in CNodes
 - now, in NIFs, as in CNodes, defininition of State in header file is no longer required. Instead of this, developer is able do decide in .spec.exs file whether to define and use State or not
 - Elixir API for communication with Unifex CNodes
 
For now, only thing, that is available in native API in NIFs, but not in CNodes, is payload handling

c_src/unifex/cnode/unifex/unifex_cnode.c
```diff
@@ -0,0 +1,228 @@
...
+
+  const int queue_size = 5;
+  if (listen(fd, queue_size)) {
+    return 1;
+  }
+
+  *listen_fd = fd;
+  return 0;
+}
+
```
bblaszkow06, line 141: Marker for me to know where to continue the review :wink:

lib/unifex/specs.ex
```diff
@@ -1,186 +1,84 @@
...
+      interface: Keyword.get(config, :interface),
+      state_type: Keyword.get(config, :state_type, nil)
+    }
   end
 
-  # Embeds code in a `quote` block. Useful when willing to store the code and parse
-  # it in runtime instead of compile time.
-  defp enquote(value) do
-    {:quote, [], [[do: value]]}
+  # Returns clear __ENV__ with proper functions/macros imported. Useful for invoking
```
bblaszkow06, line 70:
```suggestion
  # Returns a clean __ENV__ with proper functions/macros imported. Useful for invoking
```

lib/unifex/specs_dsl.ex
```diff
@@ -0,0 +1,192 @@
+defmodule Unifex.Specs.DSL do
```
bblaszkow06, line 1: I really like the change in module names :+1:

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,156 @@
+defmodule Unifex.UnifexCNode do
+  @doc """
+  Wraps Bundlex.CNode functionalities, in due to support specific Unifex's CNode behaviours
```
bblaszkow06, line 3:
```suggestion
  Wraps Bundlex.CNode functionalities to support Unifex-specific CNode behaviours
```

lib/unifex/unifex_cnode.ex
bblaszkow06: Bump

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,156 @@
...
+  end
+
+  def cast_on_start_t(on_start) do
+    on_start
+  end
+
+  @doc """
+  Casts Bundlex.CNode struct to Unifex.UnifexCNode struct or vice versa
+  """
+  def cast_cnode(%Bundlex.CNode{server: server, node: node}) do
```
bblaszkow06, line 38: cast in Erlang world usually means something different, I'd rename it to something like `convert` or `upgrade`

lib/unifex/unifex_cnode.ex
```diff
@@ -0,0 +1,156 @@
...
+    msg = [fun_name | args] |> List.to_tuple()
+
+    unifex_cnode
+    |> cast_cnode
+    |> Bundlex.CNode.send(msg)
+  end
+
+  @doc """
+      Waits timeout miliseconds on result returned from remote call of remote UnifexCNode function.
+      Generally, when function only returns values and don't send any messages, consider use of UnifexCNode.call/3, instead of this one
```
bblaszkow06, line 140: Broken modules and function references. (in this whole module)

test_fixtures/test_project_cnode_ref_generated/example.c
```diff
@@ -0,0 +1,101 @@
+#include "example.h"
```
bblaszkow06, line 1: Something is off with catalogue structure. Shouldn't the fixtures be in test/fixtures.
Also, I think the projects are doubled: test_project_cnode occurs twice in repo, on top level and under test_fixtures

lib/unifex/code_generator/base_types/int.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Unifex.CodeGenerator.BaseTypes.Int do
+  defmodule CNode do
```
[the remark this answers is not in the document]
mat-hek: If there's no type-specific implementation, it defaults to `Unifex.CodeGenerator.Default` and it works for int with NIFs. Possibly I should make that more intuitive. int64 is not yet implemented for CNodes and I'm going to defer that until it's needed

lib/unifex/code_generator/base_types/state.ex
```diff
@@ -0,0 +1,49 @@
...
+      """
+    end
+
+    @impl BaseType
+    def generate_arg_serialize(name, _ctx) do
+      ~g"""
+      env->state = ({
+        UnifexState* unifex_state = #{name};
+        unifex_state;
+      });
```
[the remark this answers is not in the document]
mat-hek: It's to ensure that the state has the type `UnifexState`, as `env->state` is `void*`

test_fixtures/test_project_cnode_ref_generated/example.c
mat-hek: I don't see the test_project_cnode under test_fixtures... https://github.com/membraneframework/unifex/tree/cnode_support/test_fixtures

lib/unifex/code_generator/base_type.ex
```diff
@@ -0,0 +1,165 @@
...
+  """
+  def generate_arg_parse(type, name, argument, postproc_fun \\ & &1, code_generator) do
+    call(
+      type,
+      :generate_arg_parse,
+      [argument, name],
+      code_generator,
+      %{postproc_fun: postproc_fun}
+    )
+    |> postproc_fun.()
```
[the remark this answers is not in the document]
mat-hek: see 
https://github.com/membraneframework/unifex/blob/cnode_support/lib/unifex/code_generator/base_types/list.ex#L53

test_fixtures/test_project_cnode_ref_generated/example.c
bblaszkow06: It's called `test_project_cnode_ref_generated` under fixtures. Now I see the difference, but still don't get why there are 3 additional catalogues at top level (outside the `test` directory)

mat-hek: if you place another project within the test directory, the tests from that project are run when you run the tests from the original project, and they simply fail

c_src/unifex/cnode/unifex/unifex_cnode.c
```diff
@@ -0,0 +1,204 @@
...
+  ei_x_encode_atom(buff, env->node_name);
+  ei_x_encode_tuple_header(buff, 2);
+  ei_x_encode_atom(buff, msg_type);
+}
+
+void unifex_cnode_send_and_free(UnifexEnv *env, erlang_pid *pid,
+                                ei_x_buff *out_buff) {
+  ei_send(env->ei_fd, pid, out_buff->buff, out_buff->index);
+  ei_x_free(out_buff);
+  free(out_buff);
```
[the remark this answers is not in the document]
mat-hek: it's actually UNIFEX_TERM

lib/unifex/code_generator/utils.ex
```diff
@@ -0,0 +1,86 @@
...
+  use Bunch
+
+  @doc """
+  Sigil used for templating generated code.
+  """
+  @spec sigil_g(String.t(), []) :: String.t()
+  def sigil_g(content, flags) do
+    if flags != [], do: raise("unsupported flags #{inspect(flags)}")
+    content
+  end
```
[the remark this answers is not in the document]
mat-hek: - we can add some flags in the future
- it's possible to use string delimiters other than `"` and `"""`

test_fixtures/test_project_cnode_ref_generated/example.c
bblaszkow06: You can avoid that by either excluding those tests or changing `:test_paths` - https://hexdocs.pm/mix/Mix.Tasks.Test.html#module-configuration.
Aside from that, even at top level, the projects should be grouped under one dir (e.g. `test_project` -> `test_projects/nif`, `test_project_cnode -> test_proects/cnode`)
Fixtures do not contain tests so you can safely move  `test_fixtures/*` -> `test/fixtures/*`

lib/unifex/code_generator/base_types/state.ex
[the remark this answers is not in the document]
bblaszkow06: But it's not valid C, that only works cuz it's GNU extension. https://stackoverflow.com/questions/1635549/in-what-versions-of-c-is-a-block-inside-parenthesis-used-to-return-a-value-valid
```suggestion
      {
        UnifexState* unifex_state = #{name};
        env->state = unifex_state;
      };
```

test_fixtures/test_project_cnode_ref_generated/example.c
mat-hek: done

lib/unifex/code_generator/base_types/string.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Unifex.CodeGenerator.BaseTypes.String do
   @moduledoc """
-  Module implementing `Unifex.CodeGenerator.BaseType` behaviour for Unifex state.
+  Module implementing `Unifex.CodeGenerator.BaseType` behaviour for strings.
```
bblaszkow06, line 3: I think the docs for all BaseTypes should mention into what native type they are translated. In this case it is important that Elixir strings become NULL-terminated C-strings

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -1,5 +1,12 @@
 defmodule Unifex.CodeGenerator.BaseTypes.List do
+  @moduledoc """
+  Module implementing `Unifex.CodeGenerator.BaseType` behaviour for lists.
+
+  They are represented in the native code as arrays.
```
bblaszkow06, line 5: It's not exhaustive, bear in mind array size is generated as well

PR #40, opened 2020-06-16
Test

mat-hek:


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.0
 jobs:
   build:
     docker:
-      - image: membrane/bionic-membrane:latest
+      - image: membrane/membrane:latest
```
bblaszkow06, line 5: You can also add a step checking for compilation warnings

lib/unifex/interface_IO.ex
```diff
@@ -29,6 +29,11 @@ defmodule Unifex.InterfaceIO do
     File.write!("#{out_base_path}.c", source)
     File.write!("#{out_base_path}.cpp", source)
 
+    Mix.shell().cmd(
+      "clang-format -style=\"{BasedOnStyle: llvm, IndentWidth: 2}\" -i " <>
+        "#{out_base_path}.h #{out_base_path}.c #{out_base_path}.cpp"
```
bblaszkow06, line 34: It would be awesome if we could store the style in `.clang-format` file

mat-hek: you mean you want the style to be configurable? this would be problematic when comparing to reference files in tests

.circleci/config.yml
mat-hek: done

lib/unifex/interface_IO.ex
bblaszkow06: I wanted the style to be saved to a file so it could be applied to the files created by user, but I've just realized that it will land in `_generated` folder so it doesn't make sense

PR #44, opened 2020-07-11
Add support for :natives key in bundlex

mickel8:
related with membraneframework/bundlex#62

### DIFF
--- a/lib/unifex/code_generator.ex
+++ b/lib/unifex/code_generator.ex
@@ -22,37 +22,32 @@ defmodule Unifex.CodeGenerator do
 
   defp get_generator(%Specs{name: name, interface: nil}) do
     {:ok, bundlex_project} = Bundlex.Project.get()
+    config = bundlex_project.config
 
-    [:nifs, :cnodes, :natives, :libs]
-    |> Enum.find(&(bundlex_project.config |> Keyword.get(&1, []) |> Keyword.has_key?(name)))
-    |> case do
-      :nifs ->
-        Unifex.CodeGenerators.NIF
+    type =
+      [:natives, :libs]
+      |> Enum.find(&(config |> Keyword.get(&1, []) |> Keyword.has_key?(name)))
 
-      :cnodes ->
-        Unifex.CodeGenerators.CNode
+    interfaces = Keyword.get(config[type][name], :interface, [])
 
-      :natives ->
-        interfaces = bundlex_project.config[:natives][name][:interfaces]
-        get_generator(get_generator_module_name(List.first(interfaces)))
-
-      _ ->
-        Unifex.CodeGenerators.NIF
+    case interfaces do
+      [] -> Unifex.CodeGenerators.NIF
+      _ -> get_generator_module_name(List.first(interfaces))
     end
   end
 
   defp get_generator(%Specs{interface: interface}) do
-    get_generator(interface)
-  end
-
-  defp get_generator(interface) do
-    Module.concat(Unifex.CodeGenerators, interface)
+    get_generator_module_name(interface)
   end
 
   defp get_generator_module_name(interface) do
-    case interface do
-      :nif -> :NIF
-      :cnode -> :CNode
-    end
+    module_name =
+      case interface do
+        :nif -> :NIF
+        :cnode -> :CNode
+        other -> other
+      end
+
+    Module.concat(Unifex.CodeGenerators, module_name)
   end
 end
--- a/lib/unifex/code_generator.ex
+++ b/lib/unifex/code_generator.ex
@@ -23,15 +23,36 @@ defmodule Unifex.CodeGenerator do
   defp get_generator(%Specs{name: name, interface: nil}) do
     {:ok, bundlex_project} = Bundlex.Project.get()
 
-    [:nifs, :cnodes]
+    [:nifs, :cnodes, :natives, :libs]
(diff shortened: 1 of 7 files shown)


mix.exs
```diff
@@ -60,7 +60,10 @@ defmodule Unifex.MixProject do
       {:ex_doc, "~> 0.19", only: :dev, runtime: false},
       {:bunch, "~> 1.0"},
       {:shmex, "~> 0.2.0"},
-      {:bundlex, "~> 0.2.0"},
+      {:bundlex,
+       git: "https://github.com/membraneframework/bundlex.git",
```
mickel8, line 64: This probably should be changed after merging membraneframework/bundlex#64

lib/unifex/code_generator.ex
```diff
@@ -22,16 +22,32 @@ defmodule Unifex.CodeGenerator do
...
+    [:natives, :libs]
+    |> Enum.find(&(config |> Keyword.get(&1, []) |> Keyword.has_key?(name)))
+    |> (fn type ->
+          interfaces = Keyword.get(config[type][name], :interface, []) |> Bunch.listify()
+
+          case interfaces do
+            [] -> Unifex.CodeGenerators.NIF
+            _ -> get_generator(get_generator_module_name(List.first(interfaces)))
+          end
+        end).()
```
mat-hek, line 36: Instead of piping into an anonymous function call, it's better to either pipe into `case do .. end` or use `Bunch.~>`. Nevertheless, this entire expression is written as if you tried to change lines 27-28 as little as possible :P Please refactor

lib/unifex/code_generator.ex
```diff
@@ -22,16 +22,32 @@ defmodule Unifex.CodeGenerator do
...
   end
 
   defp get_generator(%Specs{interface: interface}) do
+    get_generator(interface)
+  end
+
+  defp get_generator(interface) do
```
mat-hek, line 43: This clause accepting bare interface instead of specs is confusing. Can't this just be a part of `get_generator_module_name/1`?

mickel8: done

lib/unifex/code_generator.ex
mickel8: done

lib/unifex/code_generator.ex
```diff
@@ -22,16 +22,32 @@ defmodule Unifex.CodeGenerator do
...
-    [:nifs, :cnodes]
-    |> Enum.find(&(bundlex_project.config |> Keyword.get(&1, []) |> Keyword.has_key?(name)))
-    |> case do
-      :nifs -> Unifex.CodeGenerators.NIF
-      :cnodes -> Unifex.CodeGenerators.CNode
+    type =
+      [:natives, :libs]
+      |> Enum.find(&(config |> Keyword.get(&1, []) |> Keyword.has_key?(name)))
+
+    interfaces = Keyword.get(config[type][name], :interface, [])
```
mat-hek, line 31:
```suggestion
    interfaces = [:natives, :libs] |> Enum.find_value(&get_in(config, [&1, name, :interface]))
```

lib/unifex/code_generator.ex
```diff
@@ -22,16 +22,32 @@ defmodule Unifex.CodeGenerator do
...
+
+    case interfaces do
+      [] -> Unifex.CodeGenerators.NIF
+      _ -> get_generator_module_name(List.first(interfaces))
     end
```
mat-hek, line 36: I think we should raise if there's no interface at this point. Bundlex provides NIF by default for NIFs from old API, and we dropped BC for libs anyway.

lib/unifex/loader.ex
```diff
@@ -24,7 +24,7 @@ defmodule Unifex.Loader do
         arg_names = args |> Keyword.keys() |> Enum.map(&Macro.var(&1, nil))
 
         quote do
-          defnifp unquote(wrapped_name)(unquote_splicing(arg_names))
+          defnifp(unquote(wrapped_name)(unquote_splicing(arg_names)))
```
mat-hek, line 27: This shouldn't happen as `defnifp` is listed in `locals_without_parens` in  Bundlex's formatter.exs

lib/unifex/code_generator.ex
mickel8: I agree but we didn't drop BC for libs at all. I mean we dropped only situation when lib without specified interface tries to use in deps other lib that has an interface but there is still support for libs without interfaces here: 

https://github.com/membraneframework/bundlex/blob/0c0e94394b8564f6ffa4f5c5355a0fcaff8f4a68/lib/bundlex/native.ex#L81-L89

Maybe we should drop lines 82 and 83 and remove support for libs without interfaces entirely

lib/unifex/code_generator.ex
mickel8: fixed

lib/unifex/loader.ex
mickel8: fixed

mix.exs
mickel8: done

lib/unifex/code_generator.ex
mat-hek: Libs without interfaces should be supported in bundlex, but not in unifiex

mickel8: fixed

PR #45, opened 2020-07-21
Enable generating NIF and CNode boilerplate for a single spec

### LINKED ISSUE #43: Enable generating NIF and CNode boilerplate for a single spec

mickel8:
fix #43


lib/unifex/code_generators/tie_header.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Unifex.CodeGenerators.TieHeader do
```
mat-hek, line 1: `Unifex.CodeGenerators.*` namespace is intended for generators of interfaces, implementing `Unifex.CodeGenerator` behaviour. This should be `Unifex.CodeGenerator.TieHeader` or `generate_header` should be `Unifex.CodeGenerator.generate_tie_header`

lib/unifex/code_generator.ex
```diff
@@ -12,26 +12,44 @@ defmodule Unifex.CodeGenerator do
...
+  defp get_generators(%Specs{interface: interfaces}) do
+    interfaces |> Bunch.listify() |> Enum.map(&get_generator_module_name(&1))
+  end
+
+  defp get_generator_module_name(interface) do
+    module_name =
+      case interface do
+        :nif -> :NIF
+        :cnode -> :CNode
+        other -> raise "Valid interfaces are :nif and :cnode. Passed #{other}"
```
mat-hek, line 49: Interface handling in Unifex is designed to allow implementing custom interfaces by implementing `Unifex.CodeGenerators.*` modules. That approach makes that impossible

lib/unifex/code_generators/tie_header.ex
```diff
@@ -0,0 +1,22 @@
...
+    ~g"""
+    #pragma once
+
+    #ifdef NIF
+    #include "nif/#{name}.h"
+    #endif
+
+    #ifdef CNODE
+    #include "cnode/#{name}.h"
+    #endif
```
mat-hek, line 19: these should not be hardcoded but generated basing on the interfaces used

lib/unifex/interface_IO.ex
```diff
@@ -33,12 +33,26 @@ defmodule Unifex.InterfaceIO do
...
+  def store_gitignore!(dir) do
+    out_dir_name = Path.join(dir, @generated_dir_name)
+    File.mkdir_p!(out_dir_name)
+
     File.write!(Path.join(out_dir_name, ".gitignore"), """
     *.h
-    *.c
-    *.cpp
+    nif/
+    cnode/
```
mat-hek, line 55: maybe would be enough to ignore `**/*.h`, `**/*.c` and `**/*.cpp`

lib/unifex/interface_IO.ex
```diff
@@ -20,8 +20,9 @@ defmodule Unifex.InterfaceIO do
   end
 
   def store_interface!(name, dir, code) do
-    {header, source} = code
-    out_dir_name = Path.join(dir, @generated_dir_name)
+    {header, source, interface} = code
+    interface = Unifex.Helper.get_module_string(interface) |> String.downcase()
```
mickel8, line 24: I am not sure how we want to store generated files. At this moment we have e.g.
* CNode - interface name
* BUNDLEX_CNODE - compiler flag
* cnode - directory in which generated files for this interface are stored.

Maybe we should change directory name for the same as interface (generator module name) but this doesn't look too good.

lib/unifex/code_generator/tie_header.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Unifex.CodeGenerator.TieHeader do
+  @moduledoc """
+  Generates connective header file that includes proper header based
+  on selected interface.
+  """
+
+  import Unifex.CodeGenerator.Utils, only: [sigil_g: 2]
+
+  def generate_tie_header(name, interfaces) do
```
mickel8, line 9: This function generates header file but with two new lines at the end and I couldn't get rid of one of them. Have no clue what is generating them. It looks like this 
https://github.com/membraneframework/unifex/blob/35e15a77765282961e6680ae8e37f73379ead489/test/fixtures/unified_ref_generated/example.h#L10

mat-hek: It's because of triple quotes that 'append' a newline:
```
iex(1)> """
...(1)> abc
...(1)> """
"abc\n"
```

I wouldn't care about that though. `String.trim` may help if it hampers you a lot.

lib/unifex/code_generator/tie_header.ex
```diff
@@ -0,0 +1,32 @@
...
+    interfaces
+    |> Enum.map(&generate_include(name, &1))
+    |> Enum.join("\n")
+  end
+
+  defp generate_include(name, interface) do
+    interface = Unifex.Helper.get_module_string(interface)
+
+    ~g"""
+    #ifdef BUNDLEX_#{interface |> String.upcase()}
```
mat-hek, line 27: I would make a separate callback in the CodeGenerator for that and implement in the code generators like
```
defmodule Unifex.CodeGenerators.CNode do
...
@impl true
def identification_constant(), do: "BUNDLEX_CNODE"
```

lib/tasks/compile_unifex.ex
```diff
@@ -10,8 +10,12 @@ defmodule Mix.Tasks.Compile.Unifex do
     Helper.get_source_dir()
     |> InterfaceIO.get_interfaces_specs!()
     |> Enum.each(fn {name, dir, specs_file} ->
-      code = specs_file |> Specs.parse(name) |> CodeGenerator.generate_code()
-      InterfaceIO.store_interface!(name, dir, code)
+      codes = specs_file |> Specs.parse(name) |> CodeGenerator.generate_code()
+      codes |> Enum.map(&InterfaceIO.store_interface!(name, dir, &1))
```
mat-hek, line 14:
```suggestion
      codes |> Enum.each(&InterfaceIO.store_interface!(name, dir, &1))
```

lib/unifex/code_generator/tie_header.ex
mat-hek, line 9: - since it's already in `TieHeader` module, `generate` name would be sufficient
- add specs

lib/unifex/code_generator/tie_header.ex
```diff
@@ -21,10 +22,11 @@ defmodule Unifex.CodeGenerator.TieHeader do
   end
 
   defp generate_include(name, interface) do
-    interface = Unifex.Helper.get_module_string(interface)
+    interface = inspect(interface)
+    module = Module.concat(Unifex.CodeGenerators, interface)
 
     ~g"""
-    #ifdef BUNDLEX_#{interface |> String.upcase()}
+    #ifdef #{apply(module, :identification_constant, [])}
```
mat-hek, line 29:
```suggestion
    #ifdef #{module.identification_constant()}
```

PR #47, opened 2020-07-31
Update docs

### LINKED ISSUE #46: Update guide
Update guide with information about CNodes and multiple interfaces

mickel8:
fix #46


pages/creating_unifex_natives.md
```diff
@@ -0,0 +1,167 @@
...
+touch example.spec.exs
+```
+
+You may wonder where is the `_generated/nif/example.c` and `_generated/cnode/example.c`. 
+Well, as the name suggests, it will be generated based on `example.spec.exs`!
+
+Here are the contents of `example.spec.exs`:
+
+```elixir
+module Example
```
mickel8, line 79: Not sure if this example is not too simple but on the other hand I linked some more advanced ones.

pages/creating_unifex_natives.md
```diff
@@ -0,0 +1,167 @@
...
+
+interface [NIF, CNode]
+
+spec foo() :: {:ok :: label, answer :: int}
+```
```
mickel8, line 84: Not sure if this example is not too simple but on the other hand I linked some more advanced ones.

pages/creating_unifex_natives.md
```diff
@@ -0,0 +1,167 @@
...
+In case of `CNodes`, module with `Unifex.Loader` is unnecessary. We would just do:
+```elixir
+iex(2)> require Unifex.CNode
+Unifex.CNode
+iex(3)> {:ok, cnode} = Unifex.CNode.start_link(:example)
+
+10:56:38.259 [info]  Protocol 'inet_tcp': register/listen error: econnrefused
+
+ 
+10:56:38.259 [info]  Trying to start epmd...
```
mickel8, line 149: Should this be returned?

mat-hek: The first message is logged by Erlang stdlib when it can't connect to CNode because empd is not started, so there's no way to disable it. The other is logged by Bundlex. It won't happen if epmd is started, so you can remove that from here. See https://erlang.org/doc/man/epmd.html

pages/creating_unifex_natives.md
mat-hek: Maybe add some arguments to `foo`

Issue #42, opened 2020-07-09
Implement lacking types for CNodes


mat-hek:

mickel8:
string type for function parameters will be implemented in #48

mickel8:
unsigned int will also be imeplemented in #48

mickel8:
lists in #48 too

mickel8:
Types implemented in #48 
* string
* uint
* list
* atom as function param (atom as return type has already been implemented)
* pid as return type (pid as function param has already been implemented)

mickel8:
`bool` type has been implemented in #57 

Remaining types for CNodes:
* uint64 
* int64

PR #48, opened 2020-08-04
Implement some lacking types for CNodes

mickel8:
related with membraneframework/membrane_ice_plugin#2
related with #42

In this PR following types will be implemented for CNodes:
* string
* unsigned int 
* lists

### DIFF
--- a/c_src/unifex/cnode/unifex/unifex.h
+++ b/c_src/unifex/cnode/unifex/unifex.h
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <limits.h>
 
 // required for ei.h to work
 #ifndef _REENTRANT
--- a/c_src/unifex/cnode/unifex/unifex.h
+++ b/c_src/unifex/cnode/unifex/unifex.h
@@ -4,7 +4,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <limits.h>
 
 // required for ei.h to work
 #ifndef _REENTRANT
--- a/lib/unifex/code_generator/base_types/int.ex
+++ b/lib/unifex/code_generator/base_types/int.ex
@@ -24,7 +24,8 @@ defmodule Unifex.CodeGenerator.BaseTypes.Int do
     def generate_arg_serialize(name, _ctx) do
       ~g"""
       ({
-      ei_x_encode_longlong(out_buff, (long long)#{name});
+      int tmp_int = #{name};
+      ei_x_encode_longlong(out_buff, (long long)tmp_int);
       });
       """
     end
--- a/lib/unifex/code_generator/base_types/int.ex
+++ b/lib/unifex/code_generator/base_types/int.ex
@@ -12,9 +12,9 @@ defmodule Unifex.CodeGenerator.BaseTypes.Int do
     def generate_arg_parse(argument, name, _ctx) do
       ~g"""
       ({
-        long long #{name}_longlong;
-        int result = ei_decode_longlong(#{argument}->buff, #{argument}->index, &#{name}_longlong);
-        #{name} = (int)#{name}_longlong;
+        long long tmp_longlong;
+        int result = ei_decode_longlong(#{argument}->buff, #{argument}->index, &tmp_longlong);
+        #{name} = (int)tmp_longlong;
         result;
       })
       """
@@ -24,8 +24,7 @@ defmodule Unifex.CodeGenerator.BaseTypes.Int do
     def generate_arg_serialize(name, _ctx) do
       ~g"""
       ({
-      int #{name}_int = #{name};
-      ei_x_encode_longlong(out_buff, (long long)#{name}_int);
+      ei_x_encode_longlong(out_buff, (long long)#{name});
       });
       """
     end
--- a/lib/unifex/code_generator/base_types/int.ex
+++ b/lib/unifex/code_generator/base_types/int.ex
@@ -1,6 +1,8 @@
 defmodule Unifex.CodeGenerator.BaseTypes.Int do
   @moduledoc """
   Module implementing `Unifex.CodeGenerator.BaseType` behaviour for integers.
+
+  Implemented both for NIF and CNode as function parameter as well as return type.
   """
 
   defmodule CNode do
--- a/lib/unifex/code_generator/base_types/list.ex
+++ b/lib/unifex/code_generator/base_types/list.ex
@@ -9,23 +9,36 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
   alias Unifex.CodeGenerator.BaseType
 
   @impl BaseType
+  def generate_native_type(ctx) do
+    prefix = if ctx.mode == :const, do: "const ", else: ""
+
+    [
+      "#{prefix}#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}*",
+      {"unsigned int", "_length"}
+    ]
+  end
+
+  @impl BaseType
   def generate_initialization(name, _ctx) do
     ~g<#{name} = NULL;>
+  end
+
+  @impl BaseType
+  def generate_destruction(name, ctx) do
+    ~g"""
+    if(#{name} != NULL) {
+      for(unsigned int i = 0; i < #{name}_length; i++) {
+        #{BaseType.generate_destruction(ctx.subtype, :"#{name}[i]", ctx.generator)}
+      }
+      unifex_free(#{name});
+    }
+    """
   end
 
   defmodule NIF do
     @moduledoc false
     use Unifex.CodeGenerator.BaseType
     alias Unifex.CodeGenerator.BaseType
-    @impl BaseType
-    def generate_native_type(ctx) do
-      prefix = if ctx.mode == :const, do: "const ", else: ""
-
-      [
-        "#{prefix}#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}*",
-        {"unsigned int", "_length"}
-      ]
-    end
 
     @impl BaseType
     def generate_arg_serialize(name, ctx) do
@@ -48,47 +61,30 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
     def generate_arg_parse(arg, var_name, ctx) do
       elem_name = :"#{var_name}[i]"
       len_var_name = "#{var_name}_length"
+      native_type = BaseType.generate_native_type(ctx.subtype, ctx.generator)
+      subtype = ctx.subtype
+      postproc_fun = ctx.postproc_fun
+      generator = ctx.generator
 
       ~g"""
       ({
       int get_list_length_result = enif_get_list_length(env, #{arg}, &#{len_var_name});
       if(get_list_length_result){
-        #{var_name} = enif_alloc(
-          sizeof(#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}) * #{len_var_name});
+        #{var_name} = enif_alloc(sizeof(#{native_type} * #{len_var_name});
 
         for(unsigned int i = 0; i < #{len_var_name}; i++) {
-          #{BaseType.generate_initialization(ctx.subtype, elem_name, ctx.generator)}
+          #{BaseType.generate_initialization(subtype, elem_name, generator)}
         }
 
         ERL_NIF_TERM list = #{arg};
         for(unsigned int i = 0; i < #{len_var_name}; i++) {
           ERL_NIF_TERM elem;
           enif_get_list_cell(env, list, &elem, &list);
-          #{
-        BaseType.generate_arg_parse(
-          ctx.subtype,
-          elem_name,
-          ~g<elem>,
-          ctx.postproc_fun,
-          ctx.generator
-        )
-      }
+          #{BaseType.generate_arg_parse(subtype, elem_name, ~g<elem>, postproc_fun, generator)}
         }
       }
       get_list_length_result;
       })
-      """
-    end
-
-    @impl BaseType
-    def generate_destruction(name, ctx) do
-      ~g"""
-      if(#{name} != NULL) {
-        for(unsigned int i = 0; i < #{name}_length; i++) {
-          #{BaseType.generate_destruction(ctx.subtype, :"#{name}[i]", ctx.generator)}
-        }
-        unifex_free(#{name});
-      }
       """
     end
   end
@@ -97,22 +93,13 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
     @moduledoc false
     use Unifex.CodeGenerator.BaseType
     alias Unifex.CodeGenerator.BaseType
-    @impl BaseType
-    def generate_native_type(ctx) do
-      prefix = if ctx.mode == :const, do: "const ", else: ""
-
-      [
-        "#{prefix}#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}*",
-        {"int", "_length"}
-      ]
-    end
 
     @impl BaseType
     def generate_arg_serialize(name, ctx) do
       ~g"""
       ({
         ei_x_encode_list_header(out_buff, #{name}_length);
-        for(int i = 0; i < #{name}_length; i++) {
+        for(unsigned int i = 0; i < #{name}_length; i++) {
           #{BaseType.generate_arg_serialize(ctx.subtype, :"#{name}[i]", ctx.generator)}
         }
         ei_x_encode_empty_list(out_buff);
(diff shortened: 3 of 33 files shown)


test_projects/cnode/test/example_test.exs
```diff
@@ -4,16 +4,86 @@ defmodule ExampleTest do
...
+    test_string(cnode)
+    test_list(cnode)
+    test_list_as_string(cnode)
+    test_list_of_strings(cnode)
+    test_payload(cnode)
+    test_pid(cnode)
+    test_example_message(cnode)
+
+    test_undefined_function(cnode)
+    test_wrong_arguments(cnode)
```
mickel8, line 19: Some questions here. I am not sure if this is a good idea to test it all in a single unit test. I think we should make for each type separate unit test. On the other hand doing this in that way is a little faster because we create only one CNode and initialize its state only one time.

There are no all types tested here. Should I write it in this PR?

Should I write something similar for NIFs?

test_projects/cnode/test/example_test.exs
```diff
@@ -4,16 +4,86 @@ defmodule ExampleTest do
...
+
+  def test_list_as_string(cnode) do
+    0..253 |> Enum.each(fn x ->
+      list = [x, x+1, x+2]
+      assert {:ok, list} = Unifex.CNode.call(cnode, :test_list, [list])
+    end)
+    assert {:ok, [0, 0, 0]} = Unifex.CNode.call(cnode, :test_list, [[0, 0, 0]])
+    assert {:ok, [127, 127, 127]} = Unifex.CNode.call(cnode, :test_list, [[127, 127, 127]])
+    assert {:ok, [128, 128, 128]} = Unifex.CNode.call(cnode, :test_list, [[128, 128, 128]])
+    assert {:ok, [255, 255, 255]} = Unifex.CNode.call(cnode, :test_list, [[255, 255, 255]])
```
mickel8, line 51: Probably we don't need this cases because this numbers are tested above but something tells me they should be put here too.

lib/unifex/code_generator/base_types/int.ex
```diff
@@ -12,9 +12,9 @@ defmodule Unifex.CodeGenerator.BaseTypes.Int do
     def generate_arg_parse(argument, name, _ctx) do
       ~g"""
       ({
-        long long #{name}_longlong;
-        int result = ei_decode_longlong(#{argument}->buff, #{argument}->index, &#{name}_longlong);
-        #{name} = (int)#{name}_longlong;
+        long long tmp_longlong;
+        int result = ei_decode_longlong(#{argument}->buff, #{argument}->index, &tmp_longlong);
+        #{name} = (int)tmp_longlong;
```
mickel8, line 17: This was for compatibility with lists. In case of parsing lists we can end up with `in_list[i] _longlong`

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -74,18 +81,100 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+          for(int i = 0; i < #{len_var_name}; i++) {
+            #ifdef __CHAR_UNSIGNED
+              #{elem_name} = (int)p[i];
+            #else
+              #{elem_name} = (int)p[i];
+              if(#{elem_name} < 0) {
+                #{elem_name} = #{elem_name} + 256;
+              }
+            #endif
+            }
```
mickel8, line 163: This code is for list of ints between 0 and 255. Shouldn't we generate this only when `ctx.subtype` belongs to [int, unsigned int, int 64]`? It works but it's not too clear and compiler returns warnings if parse e.g. string. (cast warnings)

mickel8:
There is also a small problem with another compiler warning. Unifex generates result functions e.g. `my_function_result_ok` with parameters with keyword `const`. In case of parsing list of strings there is a warning because we are passing `char **` argument to function with `const char ** parameter` and I am not sure what we want to do with this.
`mix test` in `test_projects/cnode` will give this warning.

lib/unifex/code_generator/base_types/list.ex
mat-hek: There's one more problem there: this code doesn't return an error if someone accidentally passes such [0..255] list to a function accepting a list of strings, what may lead to a segfault. Generating the code only for some subtypes can lead to runtime errors if we add some can-be-integer subtype. What if we just converted such problematic lists to normal ones? We could decode them with ei_decode_string, then encode with `ei_x_encode_list_header` and `ei_x_encode_ulong` and finally decode normally.

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -74,18 +81,104 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+    alias Unifex.CodeGenerator.BaseType
+    @impl BaseType
+    def generate_native_type(ctx) do
+      prefix = if ctx.mode == :const, do: "const ", else: ""
+
+      [
+        "#{prefix}#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}*",
+        {"int", "_length"}
+      ]
+    end
```
mat-hek, line 108: this is the same as for NIF, let's bring it one level up

test_projects/cnode/test/example_test.exs
mat-hek: > we create only one CNode and initialize its state only one time

exunit's setup_all should help

> Not all types are tested here. Should I write test for them in this PR?

Rather just create an issue with a list of those untested

> Should I write something similar for NIFs?

mention them in the issue

test_projects/cnode/test/example_test.exs
mat-hek: where exactly they're tested?

mickel8: Here 
https://github.com/membraneframework/unifex/blob/cef221788780157ef750571114c9031b7e62282a/test_projects/cnode/test/example_test.exs#L44-L47
as triples in form of [0, 1, 2] , [1, 2, 3] and so on. Each number between 0 and 255 will be tested.
Now I see that this code could be faster by testing [0,1,2], [3,4,5],... instead of [0,1,2], [1,2,3]...

lib/unifex/code_generator/base_types/list.ex
mickel8: The difference is in signedness. For `CNode` we need `int` because functions from `ei.h` require it (e.g. `ei_get_type`). The same situation is with `generate_destruction`

test_projects/cnode/test/example_test.exs
mat-hek: oh ok, well I believe that the lines 48-51 are enough, testing each possibility doesn't bring much more value while it has to be maintained

lib/unifex/code_generator/base_types/list.ex
mat-hek: The argument types should be the same between nifs and cnodes if we want to write code compatible with both. If some functions in ei.h need different types, we should convert/cast them

lib/unifex/code_generator/base_types/int.ex
```diff
@@ -24,8 +24,7 @@ defmodule Unifex.CodeGenerator.BaseTypes.Int do
     def generate_arg_serialize(name, _ctx) do
       ~g"""
       ({
-      int #{name}_int = #{name};
-      ei_x_encode_longlong(out_buff, (long long)#{name}_int);
+      ei_x_encode_longlong(out_buff, (long long)#{name});
```
mat-hek, line 27: this was to make sure that the argument is actually integer, not only something castable to long long

test_projects/cnode/test/example_test.exs
mickel8: Remaining tests should be implemented in #52

lib/unifex/code_generator/base_types/list.ex
mickel8: This sounds good and resolves problem with accidental segfaults but unfortunately it won't work. If we decode list using `decode_string` and then encode it using `encode_list_header` `encode_ulong` we still have to decode new list using `decode_string` :smile: 

here is code 
```c
        else if(type == ERL_STRING_EXT) {
          char *p = malloc(sizeof(char) * #{len_var_name});
          res = ei_decode_string(#{arg}->buff, #{arg}->index, p);

          ei_x_buff *buff = (ei_x_buff *)malloc(sizeof(ei_x_buff));
          ei_x_new_with_version(buff);
          ei_x_encode_list_header(buff, #{len_var_name});

          for(int i = 0; i < #{len_var_name}; i++) {
            int char_value;
            #ifdef __CHAR_UNSIGNED
              char_value = (int)p[i];
            #else
              char_value = (int)p[i];
              if(char_value < 0) {
                char_value = char_value + 256;
              }
            #endif
            printf("%d\n", char_value);
            fflush(stdout);
            ei_x_encode_ulong(buff, (unsigned int)char_value);
          }
          ei_x_encode_empty_list(buff);

          res = ei_decode_list_header(#{arg}->buff, #{arg}->index, &#{len_var_name});
          #{var_name} = malloc(sizeof(#{BaseType.generate_native_type(ctx.subtype, ctx.generator)}) * #{len_var_name});
          for (int i = 0; i < #{len_var_name}; i++) {
            unsigned long tmp_ulong;
            
            // can't do it in that way. have to use decode_string 
            res = ei_decode_ulong(#{arg}->buff, #{arg}->index, &tmp_ulong); here. 
            printf("dec: %d\n", tmp_ulong);
            fflush(stdout);
            #{elem_name} = (#{BaseType.generate_native_type(ctx.subtype, ctx.generator)})tmp_ulong;
          }
        }
        res;
      })
```

mickel8: Maybe I should just do sth like this
```c
        for(int i = 0; i < #{len_var_name}; i++) {
            #ifdef __CHAR_UNSIGNED
              #{elem_name} = (#{BaseType.generate_native_type(ctx.subtype, ctx.generator)})p[i];
            #else
              int tmp = (int)p[i]
              if(tmp < 0) {
                tmp = tmp + 256;
              }
              #{elem_name} = (#{BaseType.generate_native_type(ctx.subtype, ctx.generator)})tmp;
            #endif
            }
```
`(#{BaseType.generate_native_type(ctx.subtype, ctx.generator)})tmp` will remove compiler warnings.

lib/unifex/code_generator/base_types/unsigned.ex
```diff
@@ -25,4 +25,32 @@ defmodule Unifex.CodeGenerator.BaseTypes.Unsigned do
...
+  defmodule CNode do
+    @moduledoc false
+    use Unifex.CodeGenerator.BaseType
+    alias Unifex.CodeGenerator.BaseType
+
+    @impl BaseType
+    def generate_arg_serialize(name, _ctx) do
+      ~g"""
+      ({
+        unsigned int #{name}_uint = #{name};
```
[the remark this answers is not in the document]
mickel8: fixed this and added test

mat-hek (changes requested):

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -27,66 +57,120 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+          for(unsigned int i = 0; i < #{len_var_name}; i++) {
+            int char_value;
+            if (CHAR_MIN < 0) {
+              char_value = (int)p[i];
+              if(char_value < 0) {
+                char_value = char_value + UCHAR_MAX + 1;
+              }
+            } else {
+              char_value = (int)p[i];
+            }
```
mat-hek, line 154: I think this can be avoided if you cast `p` to `uint8 *` or `unsigned char *`

mat-hek (changes requested):

lib/unifex/code_generator/base_types/list.ex
mickel8: And I was so proud of this piece of code :cry:  Fixed :+1:

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -27,66 +59,112 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+          #{BaseType.generate_initialization(subtype, elem_name, generator)}
         }
-        enif_free(#{name});
+
+        for(unsigned int i = 0; i < #{len_var_name}; i++) {
+          #{
+        BaseType.generate_arg_parse(subtype, elem_name, "unifex_buff", postproc_fun, generator)
       }
+        }
+        res = ei_skip_term(unifex_buff->buff, unifex_buff->index);
```
mickel8, line 165: This one line can seem a little tricky. It's for omitting empty list that is encoded at list's end. Added test for it.

mat-hek: you can use `ei_decode_empty_list`

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -112,44 +112,57 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
-              }
-            #endif
+            ei_x_encode_ulong(&buff, (unsigned long)unsigned_p[i]);
           }
+          ei_x_encode_empty_list(&buff);
+
+          int version;
+          int index = 0;
+          ei_decode_version(buff.buff, &index, &version);
+          res = ei_get_type(buff.buff, &index, &type, &size);
```
mat-hek, line 144: is that needed?

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -112,44 +112,57 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+        UnifexCNodeInBuff *unifex_buff;
+        unifex_buff = malloc(sizeof(UnifexCNodeInBuff));
+        if(type == ERL_STRING_EXT) {
           char *p = malloc(sizeof(char) * #{len_var_name});
           res = ei_decode_string(#{arg}->buff, #{arg}->index, p);
-          #{var_name} = malloc(sizeof(int) * #{len_var_name});
+          unsigned char *unsigned_p = (unsigned char *)p;
+
+          ei_x_buff buff;
+          ei_x_new_with_version(&buff);
```
mat-hek, line 134: `ei_x_new` would be sufficient

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -112,44 +112,57 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+          int version;
+          int index = 0;
+          ei_decode_version(buff.buff, &index, &version);
+          res = ei_get_type(buff.buff, &index, &type, &size);
+          unifex_buff->buff = buff.buff;
+          unifex_buff->index = &index;
+        } else {
+          unifex_buff->buff = #{arg}->buff;
+          unifex_buff->index = #{arg}->index;
+        }
```
mat-hek: consider moving that into something like `unifex_cnode_string_to_list`

mickel8: I moved only part that converts string to list because it seems to me a little more readable and easier to debbug than moving this whole code?

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -27,66 +59,99 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+      ({
+        int res = 1;
+        int type;
+        int size;
+
+        ei_get_type(#{arg}->buff, #{arg}->index, &type, &size);
+        #{len_var_name} = (unsigned int) size;
+
+        UnifexCNodeInBuff *unifex_buff = (UnifexCNodeInBuff *)malloc(sizeof(UnifexCNodeInBuff));
+        unifex_buff->index = (int *)malloc(sizeof(int));
```
mat-hek, line 127: Both of these variables can be allocated on the stack. That way you wouldn't need to free unifex_buff and unifex_buff->index (the latter seems missing).

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -27,66 +59,99 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+        if(type == ERL_STRING_EXT) {
+          ei_x_buff buff = unifex_cnode_string_to_list(#{arg}, #{len_var_name});
+          unifex_buff->buff = buff.buff;
+          *unifex_buff->index = 0;
+        } else {
+          unifex_buff->buff = #{arg}->buff;
+          unifex_buff->index = #{arg}->index;
+        }
+
+        res = ei_decode_list_header(unifex_buff->buff, unifex_buff->index, &size);
```
mat-hek, line 138: assignment to `res` is not needed

c_src/unifex/cnode/unifex/unifex.h
```diff
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <limits.h>
```
mat-hek, line 7: this seems not needed anymore

pages/supported_types.md
```diff
@@ -0,0 +1,35 @@
+# Supported types
+
+Unifenosupports generating code both for usage with NIF's and CNode's.
```
mat-hek, line 3:
```suggestion
Unifex supports generating code both for usage with NIF's and CNode's. 
```

pages/supported_types.md
```diff
@@ -0,0 +1,35 @@
+# Supported types
+
+Unifenosupports generating code both for usage with NIF's and CNode's. 
+Nevertheless, some types can be not implemented yet.
+Below we present which types you can use at this moment. 
+You can also refer to [#42](https://github.com/membraneframework/unifenoissues/42) to see
```
mat-hek, line 6:
```suggestion
You can also refer to [#42](https://github.com/membraneframework/unifex/issues/42) to see 
```

mat-hek (changes requested):

lib/unifex/code_generator/base_types/list.ex
mickel8: I need to pass it  to `BaseType.generate_arg_parse()` which require pointer and this was the reason. I mean `BaseType.generate_arg_parse()` generates code with `->` not with `.`.

mat-hek: You can always obtain a pointer with `&variable`, it will be valid till the function returns

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -123,19 +123,18 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+          unifex_buff.index = &index;
         } else {
-          unifex_buff->buff = #{arg}->buff;
-          unifex_buff->index = #{arg}->index;
+          unifex_buff.buff = #{arg}->buff;
+          unifex_buff.index = #{arg}->index;
         }
-
-        res = ei_decode_list_header(unifex_buff->buff, unifex_buff->index, &size);
+        res = ei_decode_list_header(unifex_buff_ptr->buff, unifex_buff_ptr->index, &size);
```
mat-hek, line 137: remove the assignment

mat-hek (approved):

PR #49, opened 2020-08-11
Precompiler

mat-hek:


lib/unifex.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Unifex do
```
mickel8, line 1: I would place it under `lib/unifex/precompiler.ex` and name this module as `defmodule Unifex.Precompiler` but up to you

mat-hek: That way one would have to put `Unifex.Precompiler` there:
https://github.com/membraneframework/unifex/pull/49/files#diff-d4447cfb5dbff840134840900481468bR32

lib/unifex/code_generator.ex
```diff
@@ -45,9 +49,8 @@ defmodule Unifex.CodeGenerator do
   defp get_generator_module_name(interface) do
     module_name =
       case interface do
-        :nif -> :NIF
-        :cnode -> :CNode
-        other -> other
```
mickel8, line 50: This line was for user defined generators as well as for these specified in `*.spec.exs` file. Removing it we will probably support only `:nif` and `:cnode` generators specified in `bundlex.exs` :roll_eyes:

lib/unifex.ex
mickel8: Oh, I see :+1:

PR #51, opened 2020-08-11
properly resolve interface names, fix dialyzer

mkaput:


lib/tasks/compile_unifex.ex
```diff
@@ -11,9 +11,9 @@ defmodule Mix.Tasks.Compile.Unifex do
     |> InterfaceIO.get_interfaces_specs!()
     |> Enum.each(fn {name, dir, specs_file} ->
       codes = specs_file |> Specs.parse(name) |> CodeGenerator.generate_code()
-      codes |> Enum.each(&InterfaceIO.store_interface!(name, dir, &1))
-      interfaces = codes |> Enum.map(fn {_header, _source, interface} -> interface end)
-      tie_header = Unifex.CodeGenerator.TieHeader.generate_header(name, interfaces)
+      Enum.each(codes, &InterfaceIO.store_interface!(name, dir, &1))
+      generators = Enum.map(codes, &elem(&1, 2))
```
mat-hek, line 15: using elem here makes that really unclear

lib/unifex/code_generator.ex
```diff
@@ -2,55 +2,58 @@ defmodule Unifex.CodeGenerator do
...
+  @spec bundlex_interface(Bundlex.Native.interface_t()) :: Specs.interface_t()
+  defp bundlex_interface(:cnode), do: CNode
+  defp bundlex_interface(:nif), do: NIF
 
-    module = Module.concat(Unifex.CodeGenerators, module_name)
-    {interface, module}
-  end
+  @spec interface_generator(Specs.interface_t()) :: module()
+  def interface_generator(CNode), do: Unifex.CodeGenerators.CNode
+  def interface_generator(NIF), do: Unifex.CodeGenerators.NIF
```
mat-hek, line 58: this way will be problematic in case of implementing more interfaces

lib/unifex/code_generator.ex
```diff
@@ -2,55 +2,58 @@ defmodule Unifex.CodeGenerator do
...
   Behaviour for code generation.
   """
+
   alias Unifex.Specs
 
   @type code_t :: String.t()
+  @type generated_code_t :: {header :: code_t, source :: code_t, generator :: module()}
 
   @callback identification_constant() :: String.t()
+  @callback subdirectory() :: String.t()
```
mat-hek, line 12: why is using just module name insufficient?

lib/unifex/specs.ex
```diff
@@ -5,6 +5,8 @@ defmodule Unifex.Specs do
   For information on how to create such specs, see `Unifex.Specs.DSL` module.
   """
 
+  @type interface_t :: NIF | CNode
```
mat-hek, line 8: - this is rather a generator, according to the new terminology
- should be `Unifex.CodeGenerator.t`

mkaput:
Sorry for not including an explanation, but pushed this PR a bit in a rush ;P

The issue was that there was a little mess in the code on convention for naming interfaces. In some places interface names were used (`NIF`), while in other generator module names (`Unifex.CodeGenerator.NIF`). I did a refactoring to stick with generator module names globally as this was the most common case.

During this I found out that in case of missing `interface` statement in `*.spec.exs`, the interface definition from `bundlex.exs` was taken, which has a form of `[:nif, :cnode]` etc. These atoms were not translated to Unifex interface names and I fixed this bug in this PR.

lib/unifex/code_generator.ex
mkaput: because module name is mixed upper/lower case while we had directory names lowercase. I went for a callback because I think it's not elegant to enforce simply lowercasing module name suffix

I think on adding a default impl for this function though

lib/unifex/code_generator.ex
mkaput: Agree, but my reasoning was that this solution would raise a `no function clause` in easily detectable place when we catch typo-style bug, instead of having us to debug effects of `Module.concat` enated atoms

lib/unifex/specs.ex
mkaput: nope, generator is a Elixir module, while interface is an atom you pass to the `interface` call in specs

mkaput:
Added tests for bundlex.exs interface case

mkaput:
ready

lib/unifex/code_generator.ex
mat-hek: To me, it seems not elegant to couple generating code stuff with IO-related stuff. Maybe naming it more universally (`io_name` ?) would help?

lib/unifex/code_generator.ex
mat-hek: Than you can do `true = Code.ensure_loaded?(module)` on the concatenated module

lib/unifex/specs.ex
mat-hek: I thought we now have:
- `NIF`, `CNode` - modules, generators
- `:nif`, `:cnode` - interfaces
also, note that this type limits the space of possible generators to these two specified in advance, which I tried to avoid

mkaput: our API is that the user can pass `NIF` or `CNode` to the `interface` call in unifex `*.spec.exs`

lib/unifex/code_generator.ex
mkaput: That sounds good, will add this check then

lib/unifex/code_generator.ex
mkaput: Well don't code generation involve setting up proper directory structure? Especially as TieHeader **code generator** uses this field.

`io_name` seems more vague to me than `subdirectory`

mat-hek: Code generators rely on the directory structure whenever they have to, but they don't specify it. Nevertheless, I would at least name that callback in a way suggesting either that it's _the_ directory where the code lands or that it's a version of the name of generator used to do IO operations.

PR #65, opened 2020-10-02
Fix warning with unused parameter in generated code

mickel8:
When the only parameter specified in function in `*.spec.exs` file is user state then compiler warns about unused `in_buff`. This PR fixes this warning.

### DIFF
--- a/lib/unifex/code_generator/utils.ex
+++ b/lib/unifex/code_generator/utils.ex
@@ -96,4 +96,9 @@ defmodule Unifex.CodeGenerator.Utils do
     |> Enum.map(&(&1 <> ";"))
     |> Enum.join("\n")
   end
+
+  @spec generate_maybe_unused_args_statements(args :: [String.t()]) :: [String.t()]
+  def generate_maybe_unused_args_statements(args) do
+    args |> Enum.map(fn arg -> ~g<UNIFEX_MAYBE_UNUSED(#{arg});> end)
+  end
 end
--- a/lib/unifex/code_generators/cnode.ex
+++ b/lib/unifex/code_generators/cnode.ex
@@ -215,6 +215,8 @@ defmodule Unifex.CodeGenerators.CNode do
     declaration = generate_caller_function_declaration({name, args})
     exit_label = "exit_#{name}_caller"
 
+    maybe_unused_args = generate_maybe_unused_args_statements(["in_buff"]) |> Enum.join("\n")
+
     args_declaration =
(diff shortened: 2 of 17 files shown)


lib/unifex/code_generators/cnode.ex
```diff
@@ -273,6 +275,10 @@ defmodule Unifex.CodeGenerators.CNode do
     ~g"UNIFEX_TERM #{name}_caller(UnifexEnv *env, UnifexCNodeInBuff *in_buff)"
   end
 
+  defp generate_maybe_unused_args_statements(args) do
```
mat-hek, line 278: you can move that to `Unifex.CodeGenerator.Utils`

lib/unifex/code_generators/cnode.ex
```diff
@@ -215,6 +215,8 @@ defmodule Unifex.CodeGenerators.CNode do
     declaration = generate_caller_function_declaration({name, args})
     exit_label = "exit_#{name}_caller"
 
+    maybe_unused_args = generate_maybe_unused_args_statements(["in_buff"]) |> Enum.join("\n")
```
mat-hek, line 218: the join is unnecessary

lib/unifex/code_generators/nif.ex
```diff
@@ -239,6 +240,10 @@ defmodule Unifex.CodeGenerators.NIF do
     """
   end
 
+  defp generate_maybe_unused_args_statements(args) do
```
mat-hek, line 243: as above

lib/unifex/code_generator/utils.ex
```diff
@@ -96,4 +96,9 @@ defmodule Unifex.CodeGenerator.Utils do
     |> Enum.map(&(&1 <> ";"))
     |> Enum.join("\n")
   end
+
+  @spec generate_maybe_unused_args_statements(args :: [String.t()]) :: [String.t()]
+  def generate_maybe_unused_args_statements(args) do
+    args |> Enum.map(fn arg -> ~g<UNIFEX_MAYBE_UNUSED(#{arg});> end)
+  end
```
mat-hek, line 103: Now, this function seems an overkill, it's simpler just to write `UNIFEX_MAYBE_UNUSED(arg)` everywhere :P But you can leave it that way if it's ok for you

mat-hek (approved):

lib/unifex/code_generator/utils.ex
mickel8: Yeah, but it should be a little simpler to add more args as `maybe unused` in the future  :sweat_smile:

Issue #60, opened 2020-09-11
Support structs


mat-hek:
Support for structs would make it much easier to pass multiple parameters to the native functions. My proposition of the API goes as follows:

In *.spec.exs, structs could be specified like
```elixir
struct my_struct :: %My.Struct{
  id: integer,
  data: [payload],
  name: string
}
```

what would assume the existence of `My.Struct` module with the elixir struct definition, and could generate the following C struct definition:

```C
struct MyStruct {
  int id;
  UnifexPayload** data;
  char* name;
}
```

and enable passing as an argument, returning and sending the struct
```elixir
spec my_fun(my_struct) :: my_struct
sends {:my_sturct :: label, my_struct}
```

That way we could also support maps with predefined, atom keys:

```elixir
struct my_map :: %{
  id: integer,
  data: [payload],
  name: string
}
```

that would not require Elixir struct. We can also consider support for inline structs/maps:
```elixir
spec my_fun(my_struct :: %My.Struct{id: integer, data: [payload], name: string}) :: (:ok :: label)
```

that would generate

```C
struct MyFunMyStruct {
  int id;
  UnifexPayload** data;
  char* name;
}
```

cc @mickel8 @mkaput @bblaszkow06

mickel8:
For me it sounds very cool :+1: 

Btw. I am also wondering if we should consider support for plain maps (`%{}`) here. Maybe it can have some common concepts or can be implemented in the same PR :roll_eyes:

mat-hek:
Good point, it was already mentioned in #32. For sure we can support the case when a map has an in-advance specified set of atom keys, that could be treated just like struct but wouldn't require module and definition on the elixir side. Support for any maps is more complicated because there's no data type to reflect them in the C std lib.

mkaput:
The proposal sound good, minor things:

- I think we should ban dots in struct names, to avoid doing some magic in name translation
- How should allocation of lists be done when returning structs?

mat-hek:
> I think we should ban dots in struct names, to avoid doing some magic in name translation

It would enforce creating those structs in dotless modules, which would be confusing and inconvenient. What I called 'explicit naming' in the original post seems to solve that better.

> How should allocation of lists be done when returning structs?

Not sure how would it differ from allocation of any other lists...

mat-hek:
Updated description basing on the comments above

mickel8:
closed in #70

PR #70, opened 2021-03-31
Struct support

FelonEkonom:
Support for mapping C/C++ structs into Elixir structs and vice versa


lib/unifex/code_generator/base_types/struct_template.ex
```diff
@@ -0,0 +1,269 @@
...
+    Code.put_compiler_option(:ignore_module_conflict, true)
+
+    compiled_code =
+      quoted_module_code(struct_type_name, struct_module_name, struct_fields)
+      |> Code.compile_quoted()
+
+    Code.put_compiler_option(:ignore_module_conflict, ignore_module_conflicts_option)
+
+    compiled_code
+  end
```
mickel8, line 190: I am wondering if this is a good idea.

At this moment we create and compile separate module for each user defined struct. These modules are exactly the same, they differ only in name. I understand this approach becasue callbacks from `BaseType` are called based on variable type so e.g. for `int` we call calbacks from `BaseTypes.Int` for string we call callbacks from `BaseTypes.String` and so on. The problem is with struct becasue we resolve them by `struct_type_name` (if I understand it properly).

The better option would be to write one module called `struct` (instead of `struct_template`) and for each user defined struct call functions from this module.

mickel8 (changes requested):

lib/unifex/code_generator/utils.ex
```diff
@@ -76,29 +76,41 @@ defmodule Unifex.CodeGenerator.Utils do
...
+  def generate_structs_definitions(config, generator, ctx) do
+    generate(config, generator, ctx)
   end
 
   @spec generate_maybe_unused_args_statements(args :: [String.t()]) :: [String.t()]
   def generate_maybe_unused_args_statements(args) do
     args |> Enum.map(fn arg -> ~g<UNIFEX_MAYBE_UNUSED(#{arg});> end)
   end
+
+  defp generate(config, generator, mapper \\ & &1, ctx) do
```
mickel8, line 109: Maybe let's name it `do_generate`. It's quite common pattern

lib/unifex/code_generators/common.ex
```diff
@@ -0,0 +1,51 @@
...
+        code_generator,
+        ctx
+      ) do
+    struct_fields_definition =
+      struct_fields
+      |> Enum.map(fn {field_name, field_type} ->
+        BaseType.generate_declaration(field_type, field_name, code_generator, ctx)
+      end)
+      |> Enum.map(&Bunch.listify/1)
+      |> Enum.flat_map(fn x -> x end)
```
mickel8, line 34: Do we need these two lines?

lib/unifex/code_generators/common.ex
```diff
@@ -0,0 +1,51 @@
...
+        {
+          struct_alias,
+          %{
+            struct_alias: struct_alias,
+            struct_module_name: struct_module_name,
+            struct_fields: struct_fields
+          }
+        }
+      end)
+      |> Enum.into(%{})
```
mickel8, line 18: Let's use `Map.new/2` instead

lib/unifex/code_generators/common.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Unifex.CodeGenerators.Common do
```
mickel8, line 1: I would add some docs and specs for module and functions here

lib/unifex/code_generators/common.ex
```diff
@@ -0,0 +1,51 @@
...
+  import Unifex.CodeGenerator.Utils, only: [sigil_g: 2]
+  alias Unifex.CodeGenerator.BaseType
+
+  def get_ctx(specs) do
```
mickel8, line 5: I am wondering about `create_ctx` but don't insist, `get_ctx` sounds good too.

lib/unifex/specs_dsl.ex
```diff
@@ -164,6 +164,27 @@ defmodule Unifex.Specs.DSL do
...
+        ...
+      }
+
+  Struct specified in such way can be used in like any other supported type, E.g.
+
+      spec my_function(in_struct :: my_struct) :: {:ok :: label, out_struct :: my_struct}
+
+  Elixir definition of %My.Struct{} should contain every field listed in specification and is not generated by Unifex
+  """
+  defmacro struct_def(struct) do
```
mickel8, line 184: According to #60 let's rename it to `struct`

lib/unifex/code_generator/base_type.ex
```diff
@@ -99,69 +101,88 @@ defmodule Unifex.CodeGenerator.BaseType do
...
+      with %{customized_types: %{^type => type_spec}} <- ctx do
+        ctx
+        |> Map.put(:type_spec, type_spec)
+      else
+        _ -> ctx
+      end
+
     module =
-      Module.concat(Unifex.CodeGenerator.BaseTypes, type |> to_string() |> String.capitalize())
+      with %{customized_types: %{^type => _type_spec}} <- ctx do
```
mickel8, line 170: If we rename `customized_types` in ctx to e.g `struct_types` then we can do 
```elixir
if Map.has_key(ctx.struct_types, type), do: .., else: ..
```
which could be more intuitive.

test_projects/cnode/test/example_test.exs
```diff
@@ -104,6 +104,13 @@ defmodule ExampleTest do
     assert {:ok, ^pid} = Unifex.CNode.call(context[:cnode], :test_pid, [pid])
   end
 
+  test "struct", context do
```
mickel8, line 107: I would also add test for improper struct to see if error messages are correct etc. We can e.g. use struct with improper field or try to pass other type as struct. Test with recursive struct would also be very useful.

mickel8 (changes requested):
This `ctx` looks very nice :wink: 
Most comments are related to formatting or naming changes

lib/unifex/specs_dsl.ex
FelonEkonom: `struct` is name of already existing function in `Kernel` module

https://hexdocs.pm/elixir/Kernel.html#struct/1

lib/unifex/code_generator/base_type.ex
FelonEkonom: After adding to the context informations about new customized types, other than structures, this place in code would require change again

lib/unifex/code_generator/base_type.ex
```diff
@@ -99,69 +101,88 @@ defmodule Unifex.CodeGenerator.BaseType do
...
 
     gen_aware_module = Module.concat(module, code_generator)
 
-    default_gen_aware_module =
-      Module.concat(Unifex.CodeGenerator.BaseTypes.Default, code_generator)
+    default_gen_aware_module = Module.concat(BaseTypes.Default, code_generator)
 
     args =
-      args ++ [Map.merge(%{generator: code_generator, type: full_type, subtype: subtype}, ctx)]
+      args ++ [Map.merge(ctx, %{generator: code_generator, type: full_type, subtype: subtype})]
```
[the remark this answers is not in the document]
FelonEkonom: Old version of code was working properly, because contexts hadn't been passed in every call of public functions from `BaseType` module. Since I have changed it, it's necessary to change this order, in due to override existing values in `ctx` - before my changes `ctx` hadn't contain these keys

test_projects/cnode/test/example_test.exs
mickel8: Ahh right, `def_struct` has been already taken too. I am not convinced about `struct_def` as it breaks convention (we probably always have `def_xxx`). Maybe `def_c_struct` or `def_unifex_struct`? @mat-hek do you have any ideas?

lib/unifex/specs_dsl.ex
mickel8: Ahh right, `def_struct` has been already taken too. I am not convinced about `struct_def` as it breaks convention (we probably always have `def_xxx`). Maybe `def_c_struct` or `def_unifex_struct`? @mat-hek do you have any ideas?

lib/unifex/code_generator/base_type.ex
mickel8: Yeah I know but at this moment I can see only structs and enums. Don't think we would have more. If we don't want to change this I would rename `customized_types` to `user_types`?

lib/unifex/code_generators/common.ex
FelonEkonom: Yes, because some of the elements of mapped list are lists too (e.g. when type of field of struct is `:list`)

lib/unifex/specs_dsl.ex
mat-hek: Maybe we could just have `type` use it also for enums? It would reflect elixir `@type` like `spec` reflects `@spec`

test_projects/cnode/c_src/example/example.spec.exs
```diff
@@ -44,3 +44,10 @@ struct_def(
 )
 
 spec test_my_struct(in_struct :: my_struct) :: {:ok :: label, out_struct :: my_struct}
+
+type outer_struct :: %Outer.Struct{
+  nested_struct: my_struct,
+  id: int
+}
```
mickel8, line 51: Maybe
```suggestion
type nested_struct :: %Nested.Struct{
  inner_struct: my_struct,
  id: int
}
```
and then we would have `test_nested_struct(in_struct :: nested_struct)`

`outer_struct` at the first glance is not too intuitive

test_projects/cnode/test/example_test.exs
```diff
@@ -109,6 +109,14 @@ defmodule ExampleTest do
 
     my_struct = %My.Struct{id: 1, name: "Jan Kowlaski", data: [1, 2, 3, 4, 5, 6, 7, 8, 9]}
     assert {:ok, ^my_struct} = Unifex.CNode.call(cnode, :test_my_struct, [my_struct])
+
+    outer_struct = %Outer.Struct{id: 2, nested_struct: my_struct}
+    assert {:ok, ^outer_struct} = Unifex.CNode.call(cnode, :test_outer_struct, [outer_struct])
+
+    invalid_struct = %Outer.Struct{id: 3, nested_struct: "Ala ma kota"}
```
mickel8, line 116: I would change "Ala ma kota" to something more universal e.g. `unifex`.

test_projects/nif/c_src/example/example.spec.exs
```diff
@@ -33,3 +33,10 @@ struct_def(
 )
```
mickel8, line 34: don't format this file so we can have
```elixir
type my_struct :: %My.Struct{
    id: int,
    data: [int],
```

test_projects/nif/test/example_test.exs
```diff
@@ -42,8 +42,16 @@ defmodule ExampleTest do
...
 
-  test "my_struct" do
+  test "struct" do
     my_struct = %My.Struct{id: 1, name: "Jan Kowlaski", data: [1, 2, 3, 4, 5, 6, 7, 8, 9]}
     assert {:ok, ^my_struct} = Example.test_my_struct(my_struct)
+
+    outer_struct = %Outer.Struct{id: 2, nested_struct: my_struct}
+    assert {:ok, ^outer_struct} = Example.test_outer_struct(outer_struct)
+
+    invalid_struct = %Outer.Struct{id: 3, nested_struct: "Ala ma kota"}
```
mickel8, line 52: As above

mickel8 (changes requested):
Some comments from previous PR have survived :wink:

mickel8 (approved):

PR #71, opened 2021-03-31
Fix NIF modules not being recompiled on spec change

FelonEkonom:


mat-hek (approved):
Seems ok, but I renamed the PR. Please rename the commit accordingly.

PR #72, opened 2021-04-14
Enum c support

### LINKED ISSUE #61: Support C enums
From the elixir side, passing/returning enums could work the same way as passing atoms, but those atoms could be converted to C enums. The spec could look like:

```elixir
spec my_fun(some_argument :: :something | :something_else, :foo | :bar) :: :some_result | :another_result
sends {:some_message, status :: :status_x | :status_y}
```

that would generate
```C
enum MyFunSomeArgumentEnum{SOMETHING, SOMETHING_ELSE};
enum MyFunArgument2Enum{FOO, BAR};
enum MyFunResultEnum{SOME_RESULT, ANOTHER_RESULT};
enum SendSomeMessageStatusEnum{STATUS_X | STATUS_Y};
```

We could also allow specifying enums explicitly for multiple uses

```elixir
enum my_enum :: :option_one | :option_two | :option_three

spec my_fun(my_enum) :: my_enum
sends {:some_message, my_enum}
```

and the C side

```C
enum MyEnum{OPTION_ONE, OPTION_TWO, OPTION_THREE};
```

cc @mickel8 @mkaput @bblaszkow06

FelonEkonom:
Support for mapping C/C++ enums into Elixir atoms and vice versa

closes #61


lib/unifex/specs_dsl.ex
```diff
@@ -181,10 +181,14 @@ defmodule Unifex.Specs.DSL do
 
   Elixir definition of %My.Struct{} should contain every field listed in specification and is not generated by Unifex
   """
-  defmacro type(struct) do
+  defmacro type({:"::", _, [_, {:%, _, _}]} = struct) do
```
mat-hek, line 184: Please update the doc, and also extend the example in the moduledoc so that it contained a struct and an enum

mat-hek (changes requested):

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -0,0 +1,184 @@
...
+      if_statements =
+        ctx.type_spec.types
+        |> Enum.map(&Atom.to_string/1)
+        |> Enum.map(fn type ->
+          if_condition =
+            if type != last_type do
+              ~g"if (#{name} == #{type |> String.upcase()})"
+            else
+              ""
+            end
```
mickel8, line 121: As far as I understand this is to have `else` instead of `else if` at the end. I would write it as 
```elixir
if_condition = ~g"if (#{name} == #{String.upcase(type)})"
```
We will have 
```c
if ()
else if()
else if()
...
else if()
```

instead of
```c
```c
if ()
else if()
else if()
...
else
```
 
but the code is much clearer

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -0,0 +1,184 @@
...
+        #{
+        BaseType.generate_arg_parse(
+          :atom,
+          :enum_as_string,
+          arg,
+          postproc_fun,
+          generator,
+          ctx
+        )
+      }
```
mickel8, line 175: As above. Shouldn't we free enum_as_string somehow?

test_projects/cnode/test/example_test.exs
```diff
@@ -119,6 +119,14 @@ defmodule ExampleTest do
     end
   end
 
+  test "enum", context do
+    cnode = context[:cnode]
+
+    assert {:ok, :option_one} = Unifex.CNode.call(cnode, :test_my_enum, [:option_one])
+    assert {:ok, :option_two} = Unifex.CNode.call(cnode, :test_my_enum, [:option_two])
+    assert {:ok, :option_three} = Unifex.CNode.call(cnode, :test_my_enum, [:option_three])
+  end
```
mickel8, line 128: We should also test some wrong atoms

test_projects/nif/test/example_test.exs
```diff
@@ -54,4 +54,10 @@ defmodule ExampleTest do
       Example.test_nested_struct(invalid_struct)
     end
   end
+
+  test "enum" do
+    assert {:ok, :option_one} = Example.test_my_enum(:option_one)
+    assert {:ok, :option_two} = Example.test_my_enum(:option_two)
+    assert {:ok, :option_three} = Example.test_my_enum(:option_three)
+  end
```
mickel8, line 62: as above

test_projects/cnode/c_src/example/example.spec.exs
```diff
@@ -49,3 +49,7 @@ type nested_struct :: %Nested.Struct{
 }
 
 spec test_nested_struct(in_struct :: nested_struct) :: {:ok :: label, out_struct :: nested_struct}
+
+type my_enum :: :option_one | :option_two | :option_three | :option_four | :option_five
+
+spec test_my_enum(in_enum :: my_enum) :: {:ok :: label, out_enum :: my_enum}
```
mickel8, line 55: Is it possible to write spec like

```elixir
spec test_my_enum(in_enum :: my_enum) ":: {:error :: label, :not_found}
```
I mean, can we write atoms directly in specs as stated in #61?

If no is it easy to introduce this? We can also make an issue for this.

mickel8 (changes requested):

lib/unifex/code_generator/base_types/enum.ex
FelonEkonom: First option generates warinings during compilation, so I decided to use second one

test_projects/cnode/c_src/example/example.spec.exs
FelonEkonom: I would create separate issue for this conception, to have a place to discuss some doubts, like how to access these inlined atoms/enums from code in C/C++

lib/unifex/code_generator/base_types/enum.ex
mickel8: Ah I see. So maybe let's try to write something like that
```elixir
    @impl BaseType
    def generate_arg_serialize(name, ctx) do
      {last_type, types} = List.pop_at(ctx.type_spec.types, -1)
      last_type = Atom.to_string(last_type)

      if_statements =
        types
        |> Enum.map(&Atom.to_string/1)
        |> Enum.map(fn type ->
          ~g"""
          if (#{name} == #{type |> String.upcase()}) {
            #{do_serialize(type, ctx)}
          }
          """
        end)
        if_statements = if_statements ++ ["{ #{do_serialize(last_type, ctx)} }"] |> Enum.join(" else ")


      ~g"""
      ({
        #{if_statements}
      });
      """
    end

    defp do_serialize(type, ctx) do
      ~g"""
      char* enum_as_string = "#{type}";
            #{
        BaseType.generate_arg_serialize(
          :atom,
          :enum_as_string,
          ctx.generator,
          ctx
        )
      }
      """
    end
```

Is it clearer?

test_projects/cnode/c_src/example/example.spec.exs
mickel8: Sure, so please create a new issue for this and link it here ;)

mickel8: Ping

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -92,6 +92,8 @@ defmodule Unifex.CodeGenerator.BaseTypes.Enum do
 
         #{if_statements}
 
+        unifex_free((void *) enum_as_string);
```
mickel8, line 95: Sorry I didn't notice it before. This way we won't free memory when `BaseType.generate_arg_parse` fails. If `BaseType.generate_arg_parse` fails it will goto exit label and it will omit this unifex_free. I am not sure how to fix it. Should we write parsing atom directly here instead of using `BaseType.generate_arg_parse`? 
One more thing, before freeing memory we shoould check if it's not equal to NULL so
```c
char * enum_as_string = NULL;
....
if (enum_as_string != NULL) {
  unifex_free...
}
```

@mat-hek

mickel8 (changes requested):

lib/unifex/code_generator/base_types/enum.ex
mat-hek: That's right. I see two ways to solve that:
- execute `BaseTypes.Atom.generate_arg_parse` manually
- declare `enum_as_string` in `generate_initialization` (it will have to have a unique name then) and clean it up in `generate_destruction`

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -0,0 +1,169 @@
...
+              ~g"if (#{name} == #{type |> String.upcase()})"
+            else
+              ""
+            end
+
+          ~g"""
+          #{if_condition} {
+            char* enum_as_string = "#{type}";
+            res = #{
+            BaseType.generate_arg_serialize(
```
mat-hek, line 39: let's call the implementation directly here too

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -0,0 +1,169 @@
...
+        #{if_statements}
+      });
+      """
+    end
+
+    defp do_serialize(type, ctx) do
+      ~g"""
+      char* enum_as_string = "#{type}";
+      #{
+        BaseType.generate_arg_serialize(
```
mat-hek, line 127: let's call the implementation directly here too

lib/unifex/code_generator/base_types/enum.ex
```diff
@@ -0,0 +1,169 @@
+defmodule Unifex.CodeGenerator.BaseTypes.Enum do
```
mat-hek, line 1: lot's of duplicated code in this module, let's reuse some parts

mat-hek (changes requested):

test/fixtures/nif_ref_generated/nif/example.cpp
```diff
@@ -183,20 +183,23 @@ UNIFEX_TERM test_my_enum_result_ok(UnifexEnv *env, MyEnum out_enum) {
                                     if (out_enum == OPTION_ONE) {
                                       char *enum_as_string = "option_one";
                                       res = enif_make_atom(env, enum_as_string);
+
```
mickel8, line 186: It's not important but can we not generate this empty lines?

lib/unifex/specs_dsl.ex
```diff
@@ -18,18 +30,71 @@ defmodule Unifex.Specs.DSL do
...
+
+      spec my_function(in_enum :: my_enum) :: {:ok :: label, out_struct :: my_struct}
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
mickel8: My bad, they are are present :man_facepalming:

mat-hek (approved):

mickel8 (approved):

PR #73, opened 2021-04-18
remove unnecessary C compilation warnigns and update tests

FelonEkonom:
Fixing a bug based on showing unnecessary warnings from `gcc` during compilation

### DIFF
--- a/lib/unifex/code_generator/base_type.ex
+++ b/lib/unifex/code_generator/base_type.ex
@@ -92,7 +92,11 @@ defmodule Unifex.CodeGenerator.BaseType do
   Uses `type` as fallback for `c:generate_native_type/1`
 
   When mode is set to :const_unless_ptr_on_ptr, function will choose to behave like it would be set to :default or :const,
-  depending on value returned by `ptr_level(type, code_generator, ctx)`
+  depending on value returned by `ptr_level(type, code_generator, ctx)`.
+  This mode can be used in places, when in general, you want to have declaration of variable with const type, but using :const
+  mode would generate code, that would require explicit cast to avoid generating warnings during compilation - e.g. in C,
+  passing argument of type `char **` to function, that expects argument of type `char const * const *` without any explicit cast,
+  will generate such a warning
   """
   @spec generate_declaration(
           t,
@@ -165,6 +169,7 @@ defmodule Unifex.CodeGenerator.BaseType do
     call(type, :generate_native_type, [], code_generator, Map.put(ctx, :mode, mode))
   end
 
+  @spec ptr_level(t, module, ctx :: map) :: integer
   def ptr_level(type, code_generator, ctx) do
     call(type, :ptr_level, [], code_generator, ctx)
   end
--- a/lib/unifex/code_generator/base_type.ex
+++ b/lib/unifex/code_generator/base_type.ex
@@ -88,24 +88,12 @@ defmodule Unifex.CodeGenerator.BaseType do
   @spec generate_declaration(
           t,
           name :: atom,
-          mode :: :default | :const | :optional_const,
+          mode :: :default | :const | :const_if_not_ptr_on_ptr,
           module,
           ctx :: map
         ) ::
           [CodeGenerator.code_t()]
   def generate_declaration(type, name, mode \\ :default, code_generator, ctx) do
-    mode =
-      cond do
-        mode in [:default, :const] ->
-          mode
-
-        is_pointer_on_pointer_type(type, code_generator, ctx) ->
-          :default
-
-        true ->
-          :const
-      end
-
     generate_native_type(type, mode, code_generator, ctx)
     |> Bunch.listify()
     |> Enum.map(fn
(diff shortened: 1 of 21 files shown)


lib/unifex/code_generators/common.ex
```diff
@@ -57,4 +57,17 @@ defmodule Unifex.CodeGenerators.Common do
...
+    BaseType.generate_native_type(type, code_generator, ctx)
+    |> Bunch.listify()
+    |> Enum.map(fn
+      {native_type, _sufix} -> native_type
+      native_type -> native_type
+    end)
+    |> Enum.map(fn native_type -> String.replace(native_type, " ", "") end)
+    |> Enum.map(fn native_type -> String.replace(native_type, "const", "") end)
+    |> Enum.find(fn native_type -> String.match?(native_type, ~r"\*\*$") end)
+    |> is_binary()
```
mat-hek, line 71: you should rather return that information for each type separately

lib/unifex/code_generators/nif.ex
```diff
@@ -131,6 +131,16 @@ defmodule Unifex.CodeGenerators.NIF do
     ~g<UNIFEX_TERM #{name}(#{args_declarations})>
   end
 
+  defp generate_args_declarations(args, mode, ctx) do
+    Enum.flat_map(args, fn {name, type} ->
+      if Common.is_pointer_on_pointer_type(type, NIF, ctx) do
+        BaseType.generate_declaration(type, name, :default, NIF, ctx)
+      else
+        BaseType.generate_declaration(type, name, mode, NIF, ctx)
+      end
```
mat-hek, line 140: Let's do that check inside `BaseType.generate_declaration`. Add a comment describing why we do that.

mat-hek (changes requested):

lib/unifex/code_generators/common.ex
```diff
@@ -57,4 +57,17 @@ defmodule Unifex.CodeGenerators.Common do
     #endif
     """
   end
+
+  def is_pointer_on_pointer_type(type, code_generator, ctx) do
```
mat-hek, line 61: add `@spec`

mat-hek (changes requested):

lib/unifex/code_generators/nif.ex
FelonEkonom: Omiting `const` takes place in args declaration of `send` and `result` functions. Are you sure, that we want to omit `const` in every use of &BaseType.generate_declaration/5, not only in this 2 cases?

mat-hek: Good point, we need a way to provide information whether that's needed or not. Maybe a separate `mode`?

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -11,6 +11,25 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
   alias Unifex.CodeGenerator.BaseType
 
   @impl BaseType
+  def generate_native_type(ctx)
+
+  def generate_native_type(%{mode: :const_if_not_ptr_on_ptr} = ctx) do
+    ptr? =
+      BaseType.generate_native_type(ctx.subtype, ctx.mode, ctx.generator, %{ctx | mode: :default})
+      |> String.replace(" ", "")
+      |> String.match?(~r"\*$")
```
mat-hek, line 20: Maybe we should have a `ptr_level(ctx)` callback, returning the number of `*` in the type, returning 0 by default? Then we could just call it in `BaseType.generate_native_type` and check if it's `>= 2`

lib/unifex/code_generator/base_types/list.ex
```diff
@@ -11,6 +11,25 @@ defmodule Unifex.CodeGenerator.BaseTypes.List do
...
+
+    mode = if not ptr?, do: :const, else: :default
+    ctx = %{ctx | mode: mode}
+
+    [
+      "#{BaseType.generate_native_type(ctx.subtype, ctx.mode, ctx.generator, ctx)} #{
+        if not ptr?, do: "const"
+      }*",
+      {"unsigned int", "_length"}
+    ]
```
mat-hek, line 30: maybe we can just call generate_native_type recursively instead?

lib/unifex/code_generator/base_types/string.ex
```diff
@@ -11,7 +11,7 @@ defmodule Unifex.CodeGenerator.BaseTypes.String do
 
   @impl BaseType
   def generate_native_type(ctx) do
-    optional_const = if ctx.mode == :const, do: "const ", else: ""
+    optional_const = if ctx.mode in [:const, :const_if_not_ptr_on_ptr], do: "const ", else: ""
```
mat-hek, line 14: I think someone is going to forget about `:const_if_not_ptr_on_ptr` sooner or later, we need to handle that better

mat-hek (changes requested):

lib/unifex/code_generator/base_types/string.ex
FelonEkonom: The only other solution I can see, is to check, if native type is pointer on pointer, in every place, we would call `&generate_native_type/1` with `%{mode: :const_if_not_ptr_on_ptr} = ctx`, but you suggested in https://github.com/membraneframework/unifex/pull/73#discussion_r615811178 to do not do this

lib/unifex/code_generator/base_types/list.ex
FelonEkonom: If you mean, to generate `const` in native code in recursive call of `generate_native_type` - it is generated there too, but only if value of `ptr?` is set to `false`. Take a look at 3 lines of code above mentioned fragment

lib/unifex/code_generator/base_type.ex
```diff
@@ -143,6 +165,10 @@ defmodule Unifex.CodeGenerator.BaseType do
     call(type, :generate_native_type, [], code_generator, Map.put(ctx, :mode, mode))
   end
 
+  def ptr_level(type, code_generator, ctx) do
```
mat-hek, line 168: missing specs

lib/unifex/code_generator/base_type.ex
```diff
@@ -84,10 +90,26 @@ defmodule Unifex.CodeGenerator.BaseType do
   provided `name`.
 
   Uses `type` as fallback for `c:generate_native_type/1`
+
+  When mode is set to :const_unless_ptr_on_ptr, function will choose to behave like it would be set to :default or :const,
+  depending on value returned by `ptr_level(type, code_generator, ctx)`
```
mat-hek, line 95: There's still no information on why we do this

mat-hek (changes requested):

mat-hek (approved):

PR #77, opened 2021-06-30
Enums and docs update

FelonEkonom:


pages/creating_unifex_natives.md
```diff
@@ -80,12 +80,100 @@ UNIFEX_TERM foo(UnifexEnv* env, int num) {
...
+`float` | `double`
+`int` | `int`
+`int64` | `int64_t`
+`payload` | `UnifexPayload`
+`pid` | `UnifexPid`
+`state` | `UnifexState *`
+`string` | `char *`
+`uint64` | `uint64_t`
+`unsigned` | `unsigned int`
+`[int]` | `int *`, `unsigned int`
```
mat-hek, line 108: I think this list should rather be a part of documentation, not the tutorial

pages/creating_unifex_natives.md
```diff
@@ -80,12 +80,100 @@ UNIFEX_TERM foo(UnifexEnv* env, int num) {
   return foo_result_ok(env, num);
 }
 ```
+You can see, that we used function `foo_result_ok` in the implementation above. For every tuple, that might be returned from our Unifex function (in this case, it is `foo`), there will be generated implementation of result function named `[orginal_function_name]_result_[label]`, where `label` is atom with type `label` in returned tuple. Type of the first argument of this function will be `UnifexEnv*`, number and types of the rest of the arguments will depend on number and types of rest of elements in the returned tuple (see the table below).
+
 It is a very simple C code that always returns the same number it gets.
```
mat-hek, line 85: since you put a paragraph above, that sentence sounds weird

pages/creating_unifex_natives.md
```diff
@@ -80,12 +80,100 @@ UNIFEX_TERM foo(UnifexEnv* env, int num) {
   return foo_result_ok(env, num);
 }
 ```
+You can see, that we used function `foo_result_ok` in the implementation above. For every tuple, that might be returned from our Unifex function (in this case, it is `foo`), there will be generated implementation of result function named `[orginal_function_name]_result_[label]`, where `label` is atom with type `label` in returned tuple. Type of the first argument of this function will be `UnifexEnv*`, number and types of the rest of the arguments will depend on number and types of rest of elements in the returned tuple (see the table below).
```
mat-hek, line 83:
```suggestion
As you can see, we used function `foo_result_ok` in the implementation above. For every tuple, that might be returned from our Unifex function (in this case, it is `foo`), there will be generated implementation of result function named `[orginal_function_name]_result_[label]`, where `label` is an atom with type `label` in returned tuple. The type of the first argument of this function will be `UnifexEnv*`, number and types of the rest of the arguments will depend on the number and types of rest of elements in the returned tuple (see the table below).
```

Please use Grammarly to verify English.

mat-hek (changes requested):

pages/creating_unifex_natives.md
mat-hek: seems not fixed

mat-hek (approved):

PR #78, opened 2021-07-26
update bundlex

mat-hek:


test/fixtures/nif_ref_generated/nif/example.c
```diff
@@ -209,15 +209,19 @@ UNIFEX_TERM test_my_enum_result_ok(UnifexEnv *env, MyEnum out_enum) {
                                     if (out_enum == OPTION_ONE) {
                                       char *enum_as_string = "option_one";
                                       res = enif_make_atom(env, enum_as_string);
+
```
mickel8, line 212: Not so important but we probably don't want these empty lines?

mickel8 (approved):

test/fixtures/nif_ref_generated/nif/example.c
mat-hek: It's going to be tough to solve, as it seems the clang formatter doesn't remove all the empty lines, but it leaves a single one. We'd have to trim that manually when generating code, but we tend not to do that since we have the formatter... On the other hand, the formatter has the `MaxEmptyLinesToKeep` option, but we'd have to set it to 0, which would remove all the empty lines from the code

PR #79, opened 2021-08-12
Add helper functions for manipulating UnifexEnv and UnifexPid for C node

mickel8:
related with membraneframework/ex_libnice#17

### DIFF
--- a/c_src/unifex/cnode/unifex/unifex.c
+++ b/c_src/unifex/cnode/unifex/unifex.c
@@ -1,10 +1,6 @@
 #include "unifex.h"
 #include "cnode.h"
 
-void *unifex_alloc(size_t size) { return malloc(size); }
-
-void unifex_free(void *pointer) { free(pointer); }
-
 UNIFEX_TERM unifex_raise(UnifexEnv *env, const char *message) {
   ei_x_buff *out_buff = (ei_x_buff *)malloc(sizeof(ei_x_buff));
   unifex_cnode_prepare_ei_x_buff(env, out_buff, "raise");
--- a/c_src/unifex/cnode/unifex/unifex.h
+++ b/c_src/unifex/cnode/unifex/unifex.h
@@ -40,8 +40,20 @@ typedef struct UnifexCNodeContext {
   UNIFEX_TERM error;
 } UnifexEnv;
 
-void *unifex_alloc(size_t size);
-void unifex_free(void *pointer);
+static inline void *unifex_alloc(size_t size) { return malloc(size); }
+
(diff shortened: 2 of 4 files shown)


c_src/unifex/cnode/unifex/unifex.c
```diff
@@ -1,9 +1,22 @@
...
+  UnifexEnv *env = (UnifexEnv *)unifex_alloc(sizeof(UnifexEnv));
+  *env = (UnifexEnv){.node_name = calloc(sizeof(char), 256),
+                     .ei_socket_fd = ERL_ERROR,
+                     .listen_fd = -1,
+                     .reply_to = NULL,
+                     .state = NULL,
+                     .released_states = NULL,
+                     .error = NULL};
+  return env;
+}
```
mickel8, line 14: I am not sure if this makes any sense.
In fact, after creating env with `unifex_alloc_env` it doesn't contain any useful data and cannot be used for e.g. sending messages from C node to Elixir process. 

I think that we are not able to avoid conditional compilation in this case so the user code that supports both NIF and C node should look like this

```c
#ifdef BUNDLEX_NIF
state->env=unifex_alloc_env();
#else
state->env = env;
#endif
state->reply_to = *unifex_self(env, &state->reply_to);
```
instead of this:

```c
state->env = unifex_alloc_env();
state->reply_to = *unifex_self(env, &state->reply_to);
```

mat-hek: A possible solution would be to make `unifex_alloc_env` accept the original env. Then we could either copy the env or just return it (and make unifex_free_env a noop) in the c node implementation. The NIF implementation could ignore the original env.

c_src/unifex/cnode/unifex/unifex.h
```diff
@@ -40,8 +40,20 @@ typedef struct UnifexCNodeContext {
...
 
-void *unifex_alloc(size_t size);
-void unifex_free(void *pointer);
+static inline void *unifex_alloc(size_t size) { return malloc(size); }
+
+static inline void unifex_free(void *pointer) { free(pointer); }
+
+static inline UnifexEnv *unifex_alloc_env(UnifexEnv *env) { return env; }
+
+static inline void unifex_clear_env(UnifexEnv *env) { UNIFEX_UNUSED(env); }
```
mat-hek, line 49:
```suggestion
static inline void unifex_clear_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
```

we use to prefix unused variables with underscore either

mat-hek (approved):

PR #81, opened 2021-08-24
Implement thread API for C node. Change `UnifexPayload` API.

mickel8:
* remove `unifex_payload_release_ptr` 
* expect ptr to allocated `UnifexPayload` struct in `unifex_payload_alloc`
* implement thread API for C node

From now user will be forced to allocate and free `UnifexPayload` struct on its own.
Functions `unifex_payload_alloc` and `unifex_payload_release` should only be used to allocate and free data field in `UnifexPayload` struct.

### DIFF
--- a/c_src/unifex/cnode/unifex/unifex.h
+++ b/c_src/unifex/cnode/unifex/unifex.h
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <pthread.h>
 
 // required for ei.h to work
 #ifndef _REENTRANT
@@ -53,6 +54,21 @@ static inline void unifex_free_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
 static inline UnifexPid *unifex_self(UnifexEnv *env, UnifexPid *pid) {
   return (UnifexPid *) memcpy(pid, env->reply_to, sizeof(UnifexPid));
 };
+
+// Threads
+typedef pthread_t UnifexTid;
+static inline int unifex_thread_create(char *name, UnifexTid *tid, void *(*func)(void *), void *args) {
+  UNIFEX_UNUSED(name);
+  return pthread_create(tid, NULL, func, args);
+}
+
+static inline void unifex_thread_exit(void *exit_val) {
+  pthread_exit(exit_val);
+}
+
+static inline int unifex_thread_join(UnifexTid tid, void **exit_val) {
+  return pthread_join(tid, exit_val);
+}
 
 UNIFEX_TERM unifex_raise(UnifexEnv *env, const char *message);
 
--- a/c_src/unifex/cnode/unifex/unifex.h
+++ b/c_src/unifex/cnode/unifex/unifex.h
@@ -45,22 +45,12 @@ static inline void *unifex_alloc(size_t size) { return malloc(size); }
 
 static inline void unifex_free(void *pointer) { free(pointer); }
 
-/**
- * Env
- * 
- * These functions are introduced to achieve interoperability between NIF and C node.
- * They does nothing.
-*/
+// Env
 static inline UnifexEnv *unifex_alloc_env(UnifexEnv *env) { return env; }
 static inline void unifex_clear_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
 static inline void unifex_free_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
 
-/** 
- * Threads
- * 
- * To use these functions compile and link with -pthread.
- * Refer to bundlex project documentation to see how to do it.
-*/
+// Threads
 typedef pthread_t UnifexTid;
 static inline int unifex_thread_create(char *name, UnifexTid *tid, void *(*func)(void *), void *args) {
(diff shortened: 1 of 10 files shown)


FelonEkonom (approved):

c_src/unifex/cnode/unifex/unifex.h
```diff
@@ -44,12 +45,35 @@ static inline void *unifex_alloc(size_t size) { return malloc(size); }
 
 static inline void unifex_free(void *pointer) { free(pointer); }
 
+/**
+ * Env
+ * 
+ * These functions are introduced to achieve interoperability between NIF and C node.
+ * They does nothing.
```
mat-hek, line 52: They don't really do nothing, as unifex_alloc_env returns env ;) I'm not sure if we should document that behaviour as it may change

c_src/unifex/cnode/unifex/unifex.h
```diff
@@ -44,12 +45,35 @@ static inline void *unifex_alloc(size_t size) { return malloc(size); }
...
 static inline UnifexEnv *unifex_alloc_env(UnifexEnv *env) { return env; }
-
 static inline void unifex_clear_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
-
 static inline void unifex_free_env(UnifexEnv *_env) { UNIFEX_UNUSED(_env); }
 
+/** 
+ * Threads
+ * 
+ * To use these functions compile and link with -pthread.
```
mat-hek, line 61: let's add that to bundlex configuration of unifex

mat-hek (changes requested):

c_src/unifex/cnode/unifex/unifex.h
mickel8: This way we will always link with `-pthread` even when someone doesn't need it. Are w okay with this?

mat-hek: I think so, do you think that's a problem?

mickel8: Just a small overhead for memory usage. I will change it

mickel8: Done

c_src/unifex/cnode/unifex/unifex.h
mickel8: Done, but I think we should consider documenting somehow C API. In case of NIF we generate code with documentation but I am not sure if this is the best solution as to generate such a code someone need to write some other code first. I think we could move all the documentation into one place.

test_projects/cnode/c_src/example/example.c
```diff
@@ -50,12 +50,13 @@ UNIFEX_TERM test_list_with_other_args(UnifexEnv *env, int *in_list,
...
 UNIFEX_TERM test_payload(UnifexEnv *env, UnifexPayload *in_payload) {
-  UnifexPayload *out_payload =
-      unifex_payload_alloc(env, UNIFEX_PAYLOAD_BINARY, in_payload->size);
+  UnifexPayload *out_payload = (UnifexPayload *)unifex_alloc(sizeof(UnifexPayload));
+  unifex_payload_alloc(env, UNIFEX_PAYLOAD_BINARY, in_payload->size, out_payload);
   memcpy(out_payload->data, in_payload->data, out_payload->size);
   out_payload->data[0]++;
   UNIFEX_TERM result = test_payload_result_ok(env, out_payload);
   unifex_payload_release(out_payload);
+  unifex_free(out_payload);
```
mat-hek, line 59:
```suggestion
  UnifexPayload out_payload;
  unifex_payload_alloc(env, UNIFEX_PAYLOAD_BINARY, in_payload->size, &out_payload);
  memcpy(out_payload.data, in_payload->data, out_payload.size);
  out_payload.data[0]++;
  UNIFEX_TERM result = test_payload_result_ok(env, &out_payload);
  unifex_payload_release(&out_payload);
```

c_src/unifex/nif/unifex/payload.c
```diff
@@ -37,38 +37,40 @@ UNIFEX_TERM unifex_payload_to_term(ErlNifEnv *env, UnifexPayload *payload) {
...
-UnifexPayload *unifex_payload_alloc(UnifexEnv *env, UnifexPayloadType type,
-                                    unsigned int size) {
-  UnifexPayload *payload = enif_alloc(sizeof(UnifexPayload));
-  payload->size = size;
-  payload->type = type;
-  payload->owned = 1;
+int unifex_payload_alloc(UnifexEnv *env, UnifexPayloadType type,
+                                    unsigned int size, UnifexPayload *payload) {
   Shmex *p_struct;
   ShmexLibResult result;
```
FelonEkonom, line 43: This variable is named `result`, despite the fact, that it is not returned anywhere in this function, so maybe we should rename it?

mat-hek (approved):

c_src/unifex/nif/unifex/payload.c
mickel8: Yeah but this is because of struct name - `ShmexLibResult`. I would leave it as it is.

FelonEkonom (approved):

PR #85, opened 2021-12-20
Release old binary on realloc

bblaszkow06:


c_src/unifex/nif/unifex/payload.c
```diff
@@ -81,11 +81,16 @@ int unifex_payload_realloc(UnifexPayload *payload, unsigned int size) {
...
   payload->size = size;
   Shmex *shmex;
+  ErlNifBinary *old_binary;
 
   switch (payload->type) {
   case UNIFEX_PAYLOAD_BINARY:
-    payload->owned = 1;
+    old_binary = &payload->payload_struct.binary;
     res = enif_realloc_binary(&payload->payload_struct.binary, size);
+    if (payload->owned) {
```
mickel8, line 90: I would do this only if realloc was successful
```suggestion
    if (payload->owned && res) {
```   

or similarly to https://github.com/erlang/otp/blob/340aefb5cdd7aa73a3b9daa1781037334d8c7341/erts/emulator/beam/erl_nif.c#L1328

```c
if (!res) {
  return res;
}   

if (payload->owned) {
  enif_release_binary(old_binary);
}
```

mickel8 (changes requested):

c_src/unifex/nif/unifex/payload.c
bblaszkow06: The first version would cause a mem leak again as the payload data pointer would be swapped anyway.
The latter however seems like the way to go and fixes another problem. Thanks!

mickel8: Good point, the first version was incorrect :sweat_smile:

mickel8 (approved):

PR #88, opened 2022-01-23
Fix building with FreeBSD.

vanillahsu:
Hi:

I make a patch to fix unifex with FreeBSD.

could you help me to review it, 

thanks.


mat-hek:
Hi @vanillahsu, thanks for the contribution ;) Unfortunately, we don't use FreeBSD, so it will be hard to help. But the changes look good and it seems they don't break anything, so I can merge if that works for you ;)

vanillahsu:
Hi @mat-hek:

yep, it works on my box, thanks!

PR #89, opened 2022-02-25
Removing uint from the documentation

daniel-jodlos:
Having discussed it with Bartek, type uint shouldn't, and isn't, supported by Unifex. It only works by using the `BaseTypes.Default` functions and System V compatibility typedef which is not present on all systems. Namely, it is not defined on Alpine, which is what our CI runs on. Therefore, mentioning it in the documentation is confusing at best.

### DIFF
--- a/pages/supported_types.md
+++ b/pages/supported_types.md
@@ -7,11 +7,10 @@ You can also refer to [#42](https://github.com/membraneframework/unifex/issues/4
 state of work on remaining types.
 
 ## NIF
-| type    | as function parameter | as return type  |
-| ------- | :-------------------: | :-------------: |
+| type      | as function parameter | as return type  |
+| --------- | :-------------------: | :-------------: |
 | `atom`    | yes                   | yes             |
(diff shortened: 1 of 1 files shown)


pages/supported_types.md
```diff
@@ -7,11 +7,10 @@ You can also refer to [#42](https://github.com/membraneframework/unifex/issues/4
 state of work on remaining types.
 
 ## NIF
-| type    | as function parameter | as return type  |
-| ------- | :-------------------: | :-------------: |
+| type      | as function parameter | as return type  |
+| --------- | :-------------------: | :-------------: |
 | `atom`    | yes                   | yes             |
 | `bool`    | yes                   | yes             |
-| `uint`    | yes                   | yes             |
```
mickel8, line 14: This should be changed to `unsigned` instead
```suggestion
| `unsigned`    | yes                   | yes             |
```

mickel8 (changes requested):

mickel8 (approved):

PR #91, opened 2022-05-12
Prepare for release 1.0

mat-hek:


mix.exs
```diff
@@ -61,13 +67,28 @@ defmodule Unifex.MixProject do
...
+      opts
+    end
+  end
+
   defp deps do
     [
       {:bunch, "~> 1.0"},
       {:shmex, "~> 0.4.0"},
-      {:bundlex, "~> 0.5.0"},
+      {:bundlex, github: "membraneframework/bundlex", branch: "release-1.0", override: true},
```
FelonEkonom, line 88: Don't we want dependency to package on hex there?

FelonEkonom (approved):

PR #95, opened 2022-07-04
Fix struct parsing when using lists

Qizot:
There was a problem when parsing structs that came from a list. 

The code generation assumed that element name was a valid singular which is being used as a temporal variable name.
Having an element name broke the parsing phase as we were operating with arrays instead of single values.

### DIFF
--- a/lib/unifex/code_generator/base_types/struct.ex
+++ b/lib/unifex/code_generator/base_types/struct.ex
@@ -76,7 +76,11 @@ defmodule Unifex.CodeGenerator.BaseTypes.Struct do
     def generate_arg_parse(arg, var_name, ctx) do
       %{postproc_fun: postproc_fun, generator: generator} = ctx
 
-      unique_sufix = String.replace("#{var_name}", ".", "_")
+      unique_sufix =
+        "#{var_name}"
+        # replace array naming
+        |> String.replace("[i]", "")
+        |> String.replace(".", "_")
 
       fields_parsing =
         ctx.type_spec.fields
--- a/lib/unifex/code_generator/base_types/struct.ex
+++ b/lib/unifex/code_generator/base_types/struct.ex
@@ -76,22 +76,18 @@ defmodule Unifex.CodeGenerator.BaseTypes.Struct do
     def generate_arg_parse(arg, var_name, ctx) do
       %{postproc_fun: postproc_fun, generator: generator} = ctx
 
-      unique_sufix =
-        "#{var_name}"
-        # replace array naming
-        |> String.replace("[i]", "")
-        |> String.replace(".", "_")
+      unique_suffix = Unifex.CodeGenerator.Utils.sanitize_var_name("#{var_name}")
 
       fields_parsing =
         ctx.type_spec.fields
         |> Enum.map_join("\n", fn {field_name, field_type} ->
           ~g"""
-          key_#{unique_sufix} = enif_make_atom(env, "#{field_name}");
-          int get_#{field_name}_result = enif_get_map_value(env, #{arg}, key_#{unique_sufix}, &value_#{unique_sufix});
+          key_#{unique_suffix} = enif_make_atom(env, "#{field_name}");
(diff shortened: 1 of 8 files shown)


lib/unifex/code_generator/base_types/struct.ex
```diff
@@ -76,7 +76,11 @@ defmodule Unifex.CodeGenerator.BaseTypes.Struct do
     def generate_arg_parse(arg, var_name, ctx) do
       %{postproc_fun: postproc_fun, generator: generator} = ctx
 
-      unique_sufix = String.replace("#{var_name}", ".", "_")
+      unique_sufix =
+        "#{var_name}"
+        # replace array naming
+        |> String.replace("[i]", "")
+        |> String.replace(".", "_")
```
bblaszkow06, line 83: @mat-hek agreed this is really ugly (the original replace was as well)
I'd suggest 2 things:
1. modify list parsing to pass a regular var name
```elixir
    @impl true
    def generate_arg_parse(arg, var_name, ctx) do
      elem_name = :"#{var_name}_i"
      len_var_name = "#{var_name}_length"
      native_type = BaseType.generate_native_type(ctx.subtype, ctx.generator, ctx)
      %{subtype: subtype, postproc_fun: postproc_fun, generator: generator} = ctx

      ~g"""
      ({
      int get_list_length_result = enif_get_list_length(env, #{arg}, &#{len_var_name});
      if(get_list_length_result){
        #{var_name} = (#{native_type} *) enif_alloc(sizeof(#{native_type}) * #{len_var_name});

        for(unsigned int i = 0; i < #{len_var_name}; i++) {
          #{native_type} #{elem_name} = #{var_name}[i];
          #{BaseType.generate_initialization(subtype, elem_name, generator, ctx)}
        }

        ERL_NIF_TERM list = #{arg};
        for(unsigned int i = 0; i < #{len_var_name}; i++) {
          ERL_NIF_TERM elem;
          enif_get_list_cell(env, list, &elem, &list);
          #{native_type} #{elem_name} = #{var_name}[i];
          #{BaseType.generate_arg_parse(subtype, elem_name, ~g<elem>, postproc_fun, generator, ctx)}
          #{var_name}[i] = #{elem_name};
        }
      }
      get_list_length_result;
      })
      """
    end
```
  It would be best to do the same in the code, that passes something with a dot as `var_name` to get rid of the need to manipulate `var_name` 
2. create a helper to sanitize variable names replacing special chars (square brackets, dots but also `->`) with an underscore

bblaszkow06 (changes requested):

test/fixtures/nif_ref_generated/nif/example.c
```diff
@@ -410,10 +410,13 @@ static ERL_NIF_TERM export_test_list(ErlNifEnv *env, int argc,
           for (unsigned int i = 0; i < in_list_length; i++) {
             ERL_NIF_TERM elem;
             enif_get_list_cell(env, list, &elem, &list);
-            if (!enif_get_int(env, elem, &in_list[i])) {
+            int in_list_i = in_list[i];
+            if (!enif_get_int(env, elem, &in_list_i)) {
```
bblaszkow06, line 414: Would it be much work to add an example for the bugged case you've encountered?

lib/unifex/code_generator/utils.ex
```diff
@@ -15,6 +15,22 @@ defmodule Unifex.CodeGenerator.Utils do
...
+  In case of arrays `var_name[i]` results in `var_name_i`.
+  """
+  @spec sanitize_var_name(String.t()) :: String.t()
+  def sanitize_var_name(var_name) do
+    var_name =
+      var_name
+      |> String.replace(".", "_")
+      |> String.replace("->", "_")
+
+    Regex.replace(~r/\[(.*)\]/, var_name, "_\\1")
```
bblaszkow06, line 31: 1. `String.replace()` accepts a list of patterns
2. It also accepts regexes (https://hexdocs.pm/elixir/String.html#replace/4)
```suggestion
    var_name
    |> String.replace([".", "->"], "_")
    |> String.replace(~r/\[(.*)\]/, "_\\1")
```

bblaszkow06 (approved):
LGTM!

PR #98, opened 2022-08-08
Fix warns occuring in C++ tests

FelonEkonom:


test/unifex/integration_test.exs
```diff
@@ -71,6 +71,10 @@ defmodule Unifex.IntegrationTest do
     "test/fixtures/#{project}_ref_generated/#{interface}"
     |> File.ls!()
     |> Enum.each(fn ref ->
+      # File.read!("test_projects/#{project}/c_src/example/_generated/#{interface}/#{ref}")
+      # |> IO.puts()
+      # IO.puts("dupa #{project} #{interface} #{ref}")
+
```
mat-hek, line 77:
```suggestion
```

test_projects/cnode/bundlex.exs
```diff
@@ -8,13 +8,14 @@ defmodule Example.BundlexProject do
   end
 
   def natives(_platform) do
+    language = System.get_env("UNIFEX_TEST_LANG") |> String.to_atom()
     [
       example: [
         src_base: "example",
-        sources: ["example.c"],
+        sources: ["example.#{language}"],
```
mat-hek, line 15: Do C and C++ files differ from each other? If not, I'd try not to duplicate their contents

mat-hek (changes requested):

test_projects/cnode/bundlex.exs
FelonEkonom: There is no difference between these two files, but C++ compiler generates warnings on try of compiling a file with `.c` extension

mat-hek: So maybe we could create the cpp files dynamically in test?

test/unifex/integration_test.exs
```diff
@@ -2,33 +2,57 @@ defmodule Unifex.IntegrationTest do
   use ExUnit.Case, async: true
 
   test "NIF test project" do
+    generate_cpp_code("nif")
+
     test_project("nif", :nif, :c)
     test_project("nif", :nif, :cpp)
+
+    delete_cpp_code("nif")
```
mat-hek, line 10: Let's rename `test_project` to something like `do_test_project` and move this to a new `test_project` function. Also, delete_cpp_code should be run with `on_exit` https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2

mat-hek (changes requested):

test/unifex/integration_test.exs
FelonEkonom: Problem is, that `on_exit` should be called from `setup` function, which is shared between many test functions and each test generates and deletes files from different directory

mat-hek: `on_exit` can be called from within a test as well. If you don't use it, the files won't be deleted when the test fails

test/unifex/integration_test.exs
```diff
@@ -75,4 +79,12 @@ defmodule Unifex.IntegrationTest do
                File.read!("test/fixtures/#{project}_ref_generated/#{interface}/#{ref}")
     end)
   end
+
+  defp generate_cpp_code(project) do
+    System.cmd("cp", ["example.c", "example.cpp"], cd: "test_projects/#{project}/c_src/example")
```
mat-hek, line 84: Actually, you can use `File.copy!/2`

test/unifex/integration_test.exs
```diff
@@ -75,4 +79,12 @@ defmodule Unifex.IntegrationTest do
...
+  end
+
+  defp delete_cpp_code(project) do
+    System.cmd("rm", ["example.cpp"], cd: "test_projects/#{project}/c_src/example")
```
mat-hek, line 88: `File.rm`

test/unifex/integration_test.exs
mat-hek: bump

test/unifex/integration_test.exs
```diff
@@ -2,60 +2,40 @@ defmodule Unifex.IntegrationTest do
...
+    generate_cpp_code(project)
 
-    test_project("bundlex_exs", :cnode, :c)
-    test_project("bundlex_exs", :cnode, :cpp)
+    for language <- [:c, :cpp] do
+      do_test_project(project, interface, language)
+    end
 
-    delete_cpp_code("bundlex_exs")
+    delete_cpp_code(project)
```
mat-hek, line 35: ON_EXIT

mat-hek (changes requested):

mat-hek (approved):

PR #100, opened 2022-10-26
MS-267 Documentation for NIFs 

Janix4000:
Final Elixir functions, created by Unifex, can now be documented using the `@doc` decorator in the `spec` file.


test_projects/documented/c_src/example/example.spec.exs
```diff
@@ -0,0 +1,36 @@
...
+"""
+spec after_undocumented_documented_function(num :: int) :: {:ok :: label, answer :: int}
+
+@doc false
+spec undocumented_false_function(num :: int) :: {:ok :: label, answer :: int}
+
+@doc """
+Test end_file_documented_function documentation
+"""
+spec end_file_documented_function(num :: int) :: {:ok :: label, answer :: int}
```
FelonEkonom, line 36: I would rather add documentation to existing testing projects, instead of creating new project

test/unifex/integration_test.exs
```diff
@@ -33,12 +37,16 @@ defmodule Unifex.IntegrationTest do
...
+  defp run_projects_tests(project, language) do
     assert {_output, 0} =
              System.cmd("bash", ["-c", "mix test 1>&2"],
                cd: "test_projects/#{project}",
                env: [{"UNIFEX_TEST_LANG", "#{language}"}]
              )
+  end
+
+  defp do_test_project(project, interface, language) do
+    run_projects_tests(project, language)
```
FelonEkonom, line 49: If you move generating docs to already existing testing modules, this change will become unnecessary

FelonEkonom:
What about generating docs for CNodes? It would be nice to have docs also when somebody is not using NIF

FelonEkonom:
Please add a reference to related Jira ticket

test_projects/documented/test/example_test.exs
```diff
@@ -0,0 +1,38 @@
...
+    check_doc_false(:undocumented_false_function)
+    check_doc(:end_file_documented_function)
+  end
+
+  defp check_doc(fun_atom) do
+    assert get_doc(Example, fun_atom) == "Test #{fun_atom} documentation"
+  end
+
+  defp check_doc_false(fun_atom) do
+    assert get_doc(Example, fun_atom) == "false"
```
FelonEkonom, line 25: Writing `@doc "false"` should not have this same effect, as writing `@doc false`, but this line suggests, that these 2 works this same way. Have you looked at generated documentation, to see, how it looks like there?

test_projects/documented/c_src/example/example.spec.exs
Janix4000: I wanted to have separate tests for the documentation functionality. 

I needed to test many separate aspects of documentation parsing and didn't want to bloat other test projects with the junky documentation.

test_projects/documented/mix.exs
```diff
@@ -0,0 +1,31 @@
...
+    [
+      extra_applications: [:logger]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      {:unifex, path: "../.."},
+      {:elixir_sense, github: "elixir-lsp/elixir_sense", only: :test}
```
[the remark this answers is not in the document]
Janix4000: This package should not be included in the testing module.

However, I'll change this in the main `mix.exs`

test_projects/documented/test/example_test.exs
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  defp get_doc(mod, func_atom) do
+    %{docs: docs} =
+      Introspection.get_all_docs(
```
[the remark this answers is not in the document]
Janix4000: `docs/3` seems to accept only raw Elixir code. I don't really know, how to use it on already compiled function.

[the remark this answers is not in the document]
Janix4000: NVM, I used `Code` module

test_projects/documented/c_src/example/example.spec.exs
FelonEkonom: I think that it is not a good idea, to create new testing project for every new feature, that we add and implementing `@doc` is not enough separate from rest of the features in `unifex`, to give it a new testing project. Even if adding `@doc` to existing testing project made them little bit less pretty, it is better, than creating new testing project for documentation

FelonEkonom: @mat-hek what are your thoughts?

mat-hek: I'd add documentation to existing projects (at least to some functions) and add unit tests for these 'many separate aspects of documentation parsing'

mix.exs
```diff
@@ -83,10 +83,10 @@ defmodule Unifex.MixProject do
 
   defp deps do
     [
+      {:membrane_core, "~> 0.10.2"},
```
FelonEkonom, line 86: Do we need it there? Unifex shouldn't have dependency to core

lib/unifex/specs.ex
```diff
@@ -91,13 +98,37 @@ defmodule Unifex.Specs do
...
+  defp parse_docs(config) do
+    config
+    |> Enum.chunk_every(2, 1)
+    |> Enum.flat_map(fn
+      [doc: doc, function: function] ->
+        [{function, doc}]
+
+      [{:doc, _doc}, {token, _not_a_function}] ->
+        Membrane.Logger.warn(
+          "Only spec functions can be documented with `@doc` decorator. Found `#{token}`."
```
FelonEkonom, line 110:
```suggestion
          "Found @doc in file #{path_to_file}:#{line_number} that does not correspond to any function"
```

FelonEkonom (changes requested):

mix.exs
Janix4000: I cannot use our logger without it

FelonEkonom: use https://hexdocs.pm/logger/1.12.3/Logger.html

test/unifex/integration_test.exs
Janix4000: Ok, applied

FelonEkonom (approved):

PR #101, opened 2022-12-14
Format generated code only when clang-format is available

mickel8:
We shouldn't require clang-format to be available on the system. Otherwise, everyone who compiles membrane package and doesn't have clang-format, will get a warning `/usr/bin/sh: 1: clang-format: not found`.

`clang-format` is required in tests so they should remain deterministic.

### DIFF
--- a/lib/unifex/interface_IO.ex
+++ b/lib/unifex/interface_IO.ex
@@ -54,10 +54,13 @@ defmodule Unifex.InterfaceIO do
     File.write!("#{out_base_path}.c", source)
     File.write!("#{out_base_path}.cpp", source)
 
-    Mix.shell().cmd(
-      "clang-format -style=\"{BasedOnStyle: llvm, IndentWidth: 2}\" -i " <>
-        "#{out_base_path}.h #{out_base_path}.c #{out_base_path}.cpp"
-    )
+    # Format generated code only when clang-format is available
+    if Mix.shell().cmd("whereis clang-format") == 0 do
+      Mix.shell().cmd(
+        "clang-format -style=\"{BasedOnStyle: llvm, IndentWidth: 2}\" -i " <>
+          "#{out_base_path}.h #{out_base_path}.c #{out_base_path}.cpp"
+      )
+    end
 
     :ok
   end
--- a/lib/unifex/interface_IO.ex
+++ b/lib/unifex/interface_IO.ex
@@ -55,10 +55,16 @@ defmodule Unifex.InterfaceIO do
     File.write!("#{out_base_path}.cpp", source)
 
     # Format generated code only when clang-format is available
-    if Mix.shell().cmd("whereis clang-format") == 0 do
-      Mix.shell().cmd(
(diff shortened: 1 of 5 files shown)


lib/unifex/interface_IO.ex
```diff
@@ -54,10 +54,13 @@ defmodule Unifex.InterfaceIO do
     File.write!("#{out_base_path}.c", source)
     File.write!("#{out_base_path}.cpp", source)
 
-    Mix.shell().cmd(
-      "clang-format -style=\"{BasedOnStyle: llvm, IndentWidth: 2}\" -i " <>
-        "#{out_base_path}.h #{out_base_path}.c #{out_base_path}.cpp"
-    )
+    # Format generated code only when clang-format is available
+    if Mix.shell().cmd("whereis clang-format") == 0 do
```
mat-hek, line 58: It seems `whereis` always returns 0 on Mac
```
iex(1)> System.cmd("whereis", ~w(dupa))
{"dupa:\n", 0}
```

test/unifex/integration_test.exs
```diff
@@ -44,10 +44,6 @@ defmodule Unifex.IntegrationTest do
   defp do_test_project(project, interface, language) do
     run_projects_tests(project, language)
 
-    # clang-format is required to format the generated code
-    # it won't match the reference files otherwise
-    assert {_output, 0} = System.cmd("clang-format", ~w(--version))
```
mat-hek, line 49: Why move this?

mat-hek (changes requested):

test/test_helper.exs
```diff
@@ -1 +1,6 @@
-ExUnit.start()
+# clang-format is required to format the generated code
+# it won't match the reference files otherwise
+case System.cmd("clang-format", ~w(--version)) do
```
FelonEkonom, line 3: 1. `System.cmd/1` raises `%ErlangError{orginal: :enoent}`, when it can't find a command
2. Why do you use one time `$ whereis clang-format` and second time `$ clang-format` to check, if there `clang-format` is available?
3. Analogically, why do you use `System.cmd/2` in one place, to execute command, but `Mix.shell().cmd/1" in another place?

FelonEkonom (changes requested):

lib/unifex/utils.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Unifex.Utils do
+  @moduledoc false
+
+  @spec clang_format_installed?() :: boolean()
+  def clang_format_installed?() do
+    if System.find_executable("clang-format"), do: true, else: false
```
mat-hek, line 6:
```suggestion
    System.find_executable("clang-format") != nil
```

test/test_helper.exs
mickel8: Honestly speaking I don't know what is the difference between `System.cmd` and `Mix.shell().cmd()` except that `Mix` might not be available in the runtime. I moved everything to `System`

mat-hek (approved):

FelonEkonom (approved):

PR #102, opened 2022-12-16
Fix version by setting it to 1.1.0

mickel8:


mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Unifex.MixProject do
   use Mix.Project
 
-  @version "1.0.2"
+  @version "1.1.0"
```
FelonEkonom, line 4: Bump in README.md as well

FelonEkonom (changes requested):

PR #107, opened 2024-02-01
Unifex cannot properly generate functions returning nil

bartkrak:
Solves: https://github.com/membraneframework/membrane_core/issues/684


lib/unifex/code_generators/nif.ex
```diff
@@ -139,7 +139,17 @@ defmodule Unifex.CodeGenerators.NIF do
...
+    labels =
+      meta
+      |> Keyword.get_values(:label)
+      |> Enum.map(fn label ->
+        if label == nil do
+          "nil"
+        else
+          label
+        end
+      end)
```
FelonEkonom, line 152:
```suggestion
      |> Enum.map(&(if &1 == nil, do: "nil", else: &1))
```
or 
```suggestion
|> Enum.map(fn 
  nil -> "nil"
  label -> label
end)

**Review by FelonEkonom — CHANGES_REQUESTED** [review]
1) Add test cases covering this fix in test projects for NIF and CNode
2) You linked comment of the issue in the PR description instead of issue itself
3) Fix CI

File: README.md [hunk 2]
```diff
@@ -36,7 +36,7 @@ defmodule MyApp.Mixfile do
 
   defp deps() do
     [
-      {:unifex, "~> 1.1"}
+      {:unifex, "~> 1.1.1"}
```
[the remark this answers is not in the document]
FelonEkonom: or use `~>>`

lib/unifex/code_generators/nif.ex
mat-hek, line 152:
```suggestion
      |> Enum.map(&Atom.to_string/1)
```

bartkrak: this does not work: 1st argument: not an atom

mat-hek: hmm, isn't it because of https://github.com/membraneframework/unifex/pull/107/files#diff-e42f57944df3819667ac906a9e2d8bb2a04bb61a089cd070966dd1aa980c5a71R54?

lib/unifex/code_generator.ex
```diff
@@ -52,7 +52,7 @@ defmodule Unifex.CodeGenerator do
     |> Enum.map(&interface_generator/1)
   end
 
-  @spec bundlex_interface(Bundlex.Native.interface_t()) :: Specs.interface_t()
+  @spec bundlex_interface(:cnode | :nif | :port) :: CNode | NIF | Port
```
FelonEkonom, line 55:
```suggestion
  @spec bundlex_interface(Bundlex.Native.native_interface()) :: Specs.interface_t()
```

CI was failing because name of the typespec in `Bundlex` has changed, so we should adjust out code to this change, instead of eradicating reference to Bundlex typespec

test_projects/nif/c_src/example/example.spec.exs
```diff
@@ -21,9 +21,12 @@ spec test_float(in_float :: float) :: {:ok :: label, out_float :: float}
 
 spec test_int(in_int :: int) :: {:ok :: label, out_int :: int}
 
+spec test_nil() :: (nil :: label)
+
 @doc """
 test_string docs
 """
+
```
FelonEkonom, line 29:
```suggestion
```

test_projects/cnode/test/example_test.exs
```diff
@@ -8,6 +8,10 @@ defmodule ExampleTest do
     [cnode: cnode]
   end
 
+  test "nil", context do
+    assert nil == Unifex.CNode.call(context[:cnode], :test_nil, [:unifex])
```
FelonEkonom, line 12: Last argument passed to `call` is list of arguments passed to native function, so it should be `[]` instead of `[:unifex]`

FelonEkonom (changes requested):

FelonEkonom (approved):

PR #110, opened 2024-02-21
Dirty macro warning + auto arity support

bartkrak:
solves: https://github.com/membraneframework/membrane_core/issues/757

### DIFF
--- a/lib/unifex/specs.ex
+++ b/lib/unifex/specs.ex
@@ -70,15 +70,11 @@ defmodule Unifex.Specs do
     {_res, binds} = Code.eval_string(specs_code, [{:unifex_config__, []}], make_env(specs_file))
     config = binds |> Keyword.fetch!(:unifex_config__) |> Enum.reverse()
 
-    {functions_args_arity, functions_results} =
+    {functions_args, functions_results} =
       config
       |> Keyword.get_values(:function)
-      |> Enum.map(fn {name, args, results} ->
-        {{{name, args}, {name, Enum.count(args)}}, {name, results}}
-      end)
+      |> Enum.map(fn {name, args, results} -> {{name, args}, {name, results}} end)
       |> Enum.unzip()
-
-    {functions_args, functions_arity} = Enum.unzip(functions_args_arity)
 
     functions_results =
       Enum.flat_map(functions_results, fn {name, results} -> Enum.map(results, &{name, &1}) end)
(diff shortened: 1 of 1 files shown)


mat-hek:
@bartkrak could we also support passing a function without arity, that would match any arity? For example `dirty :cpu, [:my_function, :another_function]`

bartkrak:
@mat-hek It's done. Now you can use macro in two ways: 
`dirty :cpu, [:something, :another_one]` - without supplying arity, it will automatically determine correct one (assuming there is no way to define a multi-arity function in unifex, we talked about this)
`dirty :cpu, something: 0, another_one: 1` - the old way for backwards compatibility.
In both cases if you misspell name of the function or provide incorrect arity you will get a nice warning.

lib/unifex/specs.ex
```diff
@@ -70,26 +70,58 @@ defmodule Unifex.Specs do
...
     functions_results =
       Enum.flat_map(functions_results, fn {name, results} -> Enum.map(results, &{name, &1}) end)
 
     functions_docs = parse_docs(config, specs_file)
 
+    dirty_functions =
+      config
+      |> Keyword.get_values(:dirty_functions)
+      |> List.flatten()
+      |> Enum.map(fn {dirty_func, type} ->
```
mat-hek, line 92: instead of map and reject, use flat_map

lib/unifex/specs.ex
```diff
@@ -70,26 +70,58 @@ defmodule Unifex.Specs do
...
+            is_tuple(dirty_func) ->
+              dirty_func
+
+            is_atom(dirty_func) ->
+              {dirty_func, nil}
+          end
+
+        {matched_name, matched_arity} = List.keyfind(functions_arity, dirty_name, 0, {nil, nil})
+
+        if matched_name == dirty_name and (matched_arity == dirty_arity or dirty_arity == nil) do
```
mat-hek, line 104:
```suggestion
        dirty_name =
          case dirty_func do
            {name, _arity} -> name
            name -> name
          end

        {matched_name, matched_args} = List.keyfind(functions_args, dirty_name, 0, {nil, nil})

        if dirty_func in [matched_name, {matched_name, length(matched_args)}] do
```

slightly more idiomatic, and you don't need `functions_arity`

lib/unifex/specs.ex
```diff
@@ -70,26 +70,58 @@ defmodule Unifex.Specs do
...
+    {functions_args_arity, functions_results} =
       config
       |> Keyword.get_values(:function)
-      |> Enum.map(fn {name, args, results} -> {{name, args}, {name, results}} end)
+      |> Enum.map(fn {name, args, results} ->
+        {{{name, args}, {name, Enum.count(args)}}, {name, results}}
+      end)
       |> Enum.unzip()
 
+    {functions_args, functions_arity} = Enum.unzip(functions_args_arity)
```
FelonEkonom, line 81:
```suggestion
functions_arity  = Enum.map(functions_args, fn {name, args} -> {name, Enum.count(args)} end)
```
You don't have to create `functions_args_arity` variable, you can just create `function_args` the way it has been done before your changes and create `functions_arity` with a simple `Enum.map`, without second call of `Enum.unzip`, IMO it will simplify your code

FelonEkonom (approved):

lib/unifex/specs.ex
```diff
@@ -81,15 +81,40 @@ defmodule Unifex.Specs do
...
+        dirty_name =
+          case dirty_func do
+            {name, _arity} -> name
+            name -> name
+          end
+
+        {matched_name, matched_args} = List.keyfind(functions_args, dirty_name, 0, {nil, nil})
+
+        if matched_name != nil and matched_args != nil and
+             dirty_func in [matched_name, {matched_name, length(matched_args)}] do
```
bartkrak, line 98: these nil checks are required, otherwise if `List.keyfind()` doesn't match `length(matched_args) ` results in a crash

varsill (approved):

FelonEkonom (approved):

PR #111, opened 2024-02-26
support for old bugged nil return functions as well as fixed ones

bartkrak:
solves: https://github.com/membraneframework/membrane_core/issues/758

Not so pretty but it enough for a temporary fix that will be deleted someday

### DIFF
--- a/lib/unifex/code_generators/nif.ex
+++ b/lib/unifex/code_generators/nif.ex
@@ -71,9 +71,7 @@ defmodule Unifex.CodeGenerators.NIF do
      * They are automatically generated and don't need to be implemented.
      */
 
-    #{Utils.generate_functions_declarations(specs.functions_results,
-    &generate_result_function_declaration/2,
-    ctx)}
+    #{generate_functions_declarations_bugged(specs.functions_results, ctx)}
 
     /*
      * Functions that send the defined messages from Nif.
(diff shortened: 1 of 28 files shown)


lib/unifex/code_generators/nif.ex
```diff
@@ -456,4 +467,58 @@ defmodule Unifex.CodeGenerators.NIF do
...
+        ~g<UNIFEX_TERM #{[name, :result, ""] |> Enum.join("_")}(#{args_declarations})>} {
+          return #{BaseType.generate_arg_serialize(:atom, :"\"#{nil}\"", NIF, ctx)};
+        }
+        """
+      else
+        ~g<>
+      end
+    end)
+    |> Enum.filter(&(&1 != ""))
+    |> Enum.map_join("\n", & &1)
```
FelonEkonom, line 522: You can use just `Enum.join/2` here and in similar places

lib/unifex/code_generators/nif.ex
```diff
@@ -456,4 +467,58 @@ defmodule Unifex.CodeGenerators.NIF do
...
+
+        ~g<UNIFEX_TERM #{[name, :result, ""] |> Enum.join("_")}(#{args_declarations})>} {
+          return #{BaseType.generate_arg_serialize(:atom, :"\"#{nil}\"", NIF, ctx)};
+        }
+        """
+      else
+        ~g<>
+      end
+    end)
+    |> Enum.filter(&(&1 != ""))
```
FelonEkonom, line 521: Instead of combining `Enum.map/2` and `Enum.filter/2` you can just use `Enum.flat_map/2` here in similar places. It will simplify your code

lib/unifex/code_generators/nif.ex
```diff
@@ -456,4 +467,58 @@ defmodule Unifex.CodeGenerators.NIF do
...
+  # This function generates bugged version of functions declarations returning nil and should be removed in v2.0.0
+  defp generate_functions_declarations_bugged(config, ctx) do
+    config
+    |> Enum.map(fn {name, result} ->
+      {_result, meta} = generate_serialization(result, ctx)
+      args = meta |> Keyword.get_values(:arg)
+
+      args_declarations =
+        [~g<UnifexEnv* env> | generate_args_declarations(args, :const_unless_ptr_on_ptr, ctx)]
+        |> Enum.join(", ")
```
FelonEkonom, line 480: Move it inside `if` statement below

lib/unifex/code_generators/nif.ex
```diff
@@ -456,4 +467,58 @@ defmodule Unifex.CodeGenerators.NIF do
...
+        meta
+        |> Keyword.get_values(:label)
+
+      if Enum.member?(labels, "nil") do
+        ~g<UNIFEX_TERM #{[name, :result, ""] |> Enum.join("_")}(#{args_declarations})>
+      else
+        ~g<>
+      end
+    end)
+    |> List.flatten()
```
FelonEkonom, line 492: I guess this `List.flatten` is unnecessary

FelonEkonom (changes requested):
I don't see bugged functions generated in fixtures. Add test checking, if they are generated correctly (remember about `(nil :: label)` and `{nil :: label, reason :: integer}` as result specs)

lib/unifex/code_generators/nif.ex
FelonEkonom: Still unchanged

lib/unifex/code_generators/nif.ex
FelonEkonom: Unchanged

FelonEkonom (changes requested):
Please, implement or comment unsolved suggestions from previous CR round

FelonEkonom (approved):

PR #113, opened 2024-09-26
~>> -> ~>

harrisi:
:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ defmodule MyApp.Mixfile do
 
   defp deps() do
     [
-      {:unifex, "~> 1.2.0"}
+      {:unifex, "~> 1.2"}
     ]
   end
 end
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ defmodule MyApp.Mixfile do
 
   defp deps() do
     [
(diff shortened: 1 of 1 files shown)


harrisi:
The credo issue is from using the latest circle ci orb, which uses Elixir 1.17. Credo 1.7.7 has a fix, but it might make sense for the orb to use a different Elixir version? Not sure what y'all think.

As a side note, the minimum Elixir version for this and Bundlex may want to bump to Elixir 1.13, since the `req` version used by Bundlex expects Elixir 1.13.

README.md
```diff
@@ -39,7 +39,7 @@ defmodule MyApp.Mixfile do
 
   defp deps() do
     [
-      {:unifex, "~>> 1.2.0"}
+      {:unifex, "~> 1.2.0"}
```
mat-hek, line 42: It should actually be

```suggestion
      {:unifex, "~> 1.2"}
```

;)

harrisi:
Would you like me to update mix.lock in this PR or open another?

harrisi:
I opened a separate PR for now. #115

mat-hek:
Oh, it would have been enough to do that here ;) I see that some linting is still not passing, let's just merge this and I'll figure it out

PR #120, opened 2026-05-20
Allow using Unifex enums and structs in NIF/Cnode state

### LINKED ISSUE #1130: [Unifex] User-defined state can't use user-defined types
When defining the state struct, which name is defined by `Unifex.Specs.DSL.state_type/1`, it has to be defined in a header file with the name corresponding to the the `spec.exs` file. This file is included at the top of the generated header, in which all types from `spec.exs` file are defined. The consequence of this is the fact that user-defined types cannot be used as fields of the user-defined state, because the state definition always appears before user-defined type definitions.

Forward declaration of State is also not feasible, because the generated source file calls `sizeof(UnifexState)`, for which the state needs to be defined.

FelonEkonom:
closes https://github.com/membraneframework/membrane_core/issues/1130


lib/unifex/interface_IO.ex
```diff
@@ -5,6 +5,7 @@ defmodule Unifex.InterfaceIO do
 
   @spec_name_sufix ".spec.exs"
   @generated_dir_name "_generated"
+  @types_header_sufix "_types_definitions"
```
Noarkhh:
```suggestion
  @types_header_suffix "_types"
```
typo in `suffix` + I think `_types` is sufficient

Noarkhh (changes requested):

mat-hek (commented):
The goal is that this works: https://github.com/membraneframework/membrane_av1_plugin/compare/master...mf/unifex

lib/unifex/code_generators/cnode.ex
```diff
@@ -95,6 +95,29 @@ defmodule Unifex.CodeGenerators.CNode do
     """
   end
 
+  defp pargma_and_includes() do
```
Noarkhh:
```suggestion
  defp pragma_and_includes() do
```

Noarkhh (changes requested):
I found a bug where you now can't use user-defined enums in user-defined structs. It all stems from the casing - structs use snake case, and enums use camel case (we could discuss unifying that). Changes introduced in this PR make it so the following definitions in `.spec.ex` file:

```elixir
type my_enum :: :a | :b 

type my_struct :: %MyStruct{enum_field: my_enum}
```

get translated to the following C definitions:

```c
enum MyEnum_t { MY_ENUM_A, MY_ENUM_B };
typedef enum MyEnum_t MyEnum;

struct my_struct_t {
  my_enum enum_field;
};
typedef struct my_struct_t my_struct;
```

which don't work, since there's no type `my_enum`

lib/unifex/code_generators/nif.ex
```diff
@@ -96,6 +83,41 @@ defmodule Unifex.CodeGenerators.NIF do
...
+    extern "C" {
+    #endif
+
+    #{Utils.generate_enums_definitions(specs.enums,
+    &Common.generate_enum_native_definition/2,
+    %{})}
+
+    #{Utils.generate_structs_definitions(specs.structs,
+    &generate_struct_native_definition/2,
+    %{})}
```
Noarkhh: This fixes the bug

```suggestion
  def generate_types_header(specs) do
    ctx = Common.create_ctx(specs)
    
    ~g"""
    #{pargma_and_includes()}

    #ifdef __cplusplus
    extern "C" {
    #endif

    #{Utils.generate_enums_definitions(specs.enums,
    &Common.generate_enum_native_definition/2,
    ctx)}

    #{Utils.generate_structs_definitions(specs.structs,
    &generate_struct_native_definition/2,
    ctx)}
```

lib/unifex/code_generator/tie_header.ex
```diff
@@ -21,9 +20,12 @@ defmodule Unifex.CodeGenerator.TieHeader do
   end
 
   defp generate_include(name, generator) do
+    types_header_filename = Unifex.InterfaceIO.types_header_filename(name)
+
     ~g"""
     #ifdef #{generator.identification_constant()}
     #include "#{generator.interface_io_name()}/#{name}.h"
+    #include "#{generator.interface_io_name()}/#{types_header_filename}"
```
Noarkhh, line 27: My LSP complains when this include is below the other one, since the types are used inside it. I also think it makes more sense, but it's not necessary

Noarkhh (changes requested):

mat-hek (approved):

PR #123, opened 2026-07-07
Add non-blocking native logger for NIF and CNode backends

khamilowicz:
## Summary
- Adds a thread-safe, non-blocking logger (`unifex_log()`) that native NIF/CNode code can call to send log messages to the BEAM, forwarded through a new `Unifex.Logger` GenServer to Elixir's `Logger`. Works identically for both backends.
- `Unifex.Logger` is opt-in (`config :unifex, enable_logger: true`) so consumers that never log don't pay for an idle process/thread.
- Fixes the include path so any consumer can `#include <unifex/logger.h>` with no extra bundlex config, plus several native-side safety fixes (NULL message guard, `pthread_create` failure handling, retry-on-failure for the NIF fallback env).
- Adds a guide (`pages/logger.md`) and test coverage (unit tests, an integration test project, and a wired-in integration test run).

## Test plan
- [x] `mix test` (repo test suite, including new logger unit/integration tests and the `logger_test` integration project)
- [x] `mix docs` builds cleanly with the new guide
- [x] Verified end-to-end against a standalone example NIF project: config enables `Unifex.Logger`, `unifex_log()` calls are received and forwarded to `Logger` with correct formatting/metadata
- [ ] CNode backend logging exercised only via a successful build, not a live end-to-end run (would need a real `ei`-connected CNode test harness)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/unifex/logger.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  @impl true
+  def handle_info(msg, state) do
+    Logger.warning("Unifex.Logger received unknown message: #{inspect(msg)}")
+    {:noreply, state}
+  end
+
+  @doc false
+  @spec normalize_level(atom()) :: atom()
+  def normalize_level(level) when level in @valid_levels, do: level
```
FelonEkonom: Why do we have to support unnormalized log levels? What would cause having such a level?

c_src/unifex/cnode/unifex/cnode.c
```diff
@@ -21,9 +22,28 @@ void unifex_cnode_prepare_ei_x_buff(UnifexEnv *env, ei_x_buff *buff,
   ei_x_encode_atom(buff, msg_type);
 }
 
+// The socket fd is shared between the main receive loop and a background
+// logger thread (see logger_cnode.c), so every send on it must go through
+// one of these to stay serialized via socket_mutex.
+int unifex_cnode_locked_send(UnifexEnv *env, erlang_pid *pid, char *buff,
+                             int len) {
+  pthread_mutex_lock(&env->socket_mutex);
```
FelonEkonom, line 30: Correct me if I am wrong, but I thought that all we have to do from the C side is to do a proper `send` to a proper process (the same way as original Elixir `Logger` works under the hood, AFAIK).

If so, why do we need any locks or mutexes in the code?

c_src/unifex/cnode/unifex/cnode.c
```diff
@@ -232,6 +257,13 @@ int unifex_cnode_init(int argc, char **argv, UnifexEnv *env) {
 }
 
 void unifex_cnode_destroy(UnifexEnv *env) {
+  // Stop and join the logger's worker thread before tearing down the socket
+  // and mutex it may still be using to send a queued message. This is a
+  // no-op if the logger was never started (e.g. init failed before
+  // unifex_logger_cnode_init() ran) and safe to call again from the
+  // process-exit destructor (unifex_logger_cleanup() is idempotent).
```
FelonEkonom, line 264: Please, remove all unnecessary comments 😉

lib/unifex/logger.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  Not started by default. Enable it with:
+
+      config :unifex, enable_logger: true
+  """
+
+  use GenServer
+  require Logger
+
+  @router_name :"Elixir.Unifex.Logger"
```
FelonEkonom: Isn't it just `__MODULE__` here?

lib/unifex/app.ex
```diff
@@ -4,8 +4,13 @@ defmodule Unifex.App do
...
-    children = [Unifex.Counter]
+    children = [Unifex.Counter] ++ if(logger_enabled?(), do: [Unifex.Logger], else: [])
     opts = [strategy: :one_for_one, name: __MODULE__]
     Supervisor.start_link(children, opts)
   end
+
+  # The native logger queue/worker thread (see c_src/unifex/unifex/logger.c)
+  # is compiled into every unifex consumer, but only projects that actually
+  # call unifex_log() need this GenServer running to receive its messages.
+  defp logger_enabled?, do: Application.get_env(:unifex, :enable_logger, false)
```
FelonEkonom, line 15: Why did you make starting logger configurable? Why does it default to false?

lib/unifex/logger.ex
```diff
@@ -0,0 +1,79 @@
...
+  @moduledoc """
+  Generic logger for handling log messages from C NIFs via Unifex.
+
+  This process receives log messages from C code and forwards them to Elixir's Logger.
+  It can be used by any NIF library that needs to log messages to the BEAM.
+
+  Not started by default. Enable it with:
+
+      config :unifex, enable_logger: true
+  """
```
FelonEkonom: Why is this module a part of public API?

```
  This process receives log messages from C code and forwards them to Elixir's Logger.
  It can be used by any NIF library that needs to log messages to the BEAM.
```

- it might be true, but why is it mentioned in docs? if somebody wants to log something, he should just call `unifex_log` function from C, no need to sending anything explicitly to this module

README.md
```diff
@@ -49,6 +49,8 @@ end
 
 For detailed usage description see [Creating Unifex Natives](https://hexdocs.pm/unifex/creating_unifex_natives.html) guide.
 
+To send log messages from native code to Elixir's `Logger`, see the [Logging from native code](https://hexdocs.pm/unifex/logger.html) guide.
```
FelonEkonom, line 52: you link here moduledocs of `Unifex.Logger`, but it would make more sense to link `"pages/logger.md"` instead

FelonEkonom (changes requested):

lib/unifex/logger.ex
khamilowicz: User can pass any string to `unifex_log`, so it is possible to break it.

c_src/unifex/cnode/unifex/cnode.c
khamilowicz: I discussed with @varsill about drawing inspiration from [this Rust logger for NIFs](https://github.com/davydog187/nif_logger). The library uses a queue to send messages to a GenServer in an orderly way.
Since this logger is meant to log messages executed in C, we want to avoid making it a bottleneck. The main issue with the previous implementation was that looking up the GenServer process by name was extremely time-consuming and shouldn’t happen on the main thread. Caching the PID would be unreliable, as the process could terminate at any moment. This implementation processes messages on a separate thread, ensuring it doesn’t slow down the main job.

c_src/unifex/cnode/unifex/cnode.c
khamilowicz: 👍

lib/unifex/app.ex
khamilowicz: Well, if you don't use logger in the native code, why start it? Native logger is not the main feature of unifex, so it doesn't need to be obligatory.

lib/unifex/logger.ex
khamilowicz: You mean that docs should be removed, right? no problem

README.md
khamilowicz: All links in readme point to html versions of documents in `pages`. I think it will apply to this link too.

lib/unifex/logger.ex
FelonEkonom: I mean that public docs should describe public API and how our library behaves from the outside because it is what interests developers using our library. Whereas, by default, it should not be a description of how Unifex works from the inside (it may do it in cases where it make sense, in order to describe tool's outside behaviour)

mix.exs
```diff
@@ -57,6 +57,7 @@ defmodule Unifex.MixProject do
         "pages/supported_types.md",
         "pages/logger.md"
       ],
+      skip_undefined_reference_warnings_on: ["pages/logger.md"],
```
FelonEkonom, line 60: Why do we skip undefined reference warnings? Is it because of marking Logger moduledoc as false?

c_src/unifex/cnode/unifex/cnode.c
FelonEkonom: If you discussed the shape of the implementation with @varsill, I think he should be marked as a reviewer too.
BTW what is the time difference between sending message to a process by atom name vs PID? I hear about it for the first time, is it really a big deal? I am curious 🤔

lib/unifex/app.ex
FelonEkonom: Ok, but this could be an another reason to avoid spawning helper logging processes

lib/unifex/logger.ex
FelonEkonom: We could provide an enum to specify a log level. It would remove the risk of user making a typo and the need for handling invalid log levels.

mix.exs
khamilowicz: yes

c_src/unifex/unifex/logger.c
```diff
@@ -54,13 +54,22 @@ static void free_tags_copy(char **tags, unsigned int tags_length) {
...
+const char *unifex_log_level_to_string(UnifexLogLevel level) {
+  switch (level) {
+    case UNIFEX_LOG_LEVEL_DEBUG: return "debug";
+    case UNIFEX_LOG_LEVEL_INFO:  return "info";
+    case UNIFEX_LOG_LEVEL_WARN:  return "warning";
+    case UNIFEX_LOG_LEVEL_ERROR: return "error";
+    default: return "unknown";
+  }
+}
+
```
FelonEkonom, line 72: I am afraid it is not safe. You return a pointer to a place on the stack that might be overwritten after `unifex_log_level_to_string` call ends. If I am right it might stop working at any time.

c_src/unifex/unifex/logger.h
```diff
@@ -23,14 +23,19 @@ extern "C" {
...
-#define UNIFEX_LOG_LEVEL_ERROR "error"
+// Log level enum for compile-time type safety
+typedef enum {
+  UNIFEX_LOG_LEVEL_DEBUG,
+  UNIFEX_LOG_LEVEL_INFO,
+  UNIFEX_LOG_LEVEL_WARN,
+  UNIFEX_LOG_LEVEL_ERROR,
+} UnifexLogLevel;
+
+// Convert log level enum to string representation (for internal use by backends)
```
FelonEkonom: There is no reason to keep such comments in the codebase

FelonEkonom (changes requested):
Please don't force push your changes. You can always squash all commits during merge. I looked for `normalize_level` deletion in the diff, but I cannot see it, neither in `changes since your last review` nor in `all commits`. Force pushes make reviewing the PR much harder.

c_src/unifex/unifex/logger.c
khamilowicz: I don't think that's the case. Static strings should be safe to return.

https://stackoverflow.com/questions/76818467/is-it-really-safe-to-return-a-string-literal-from-a-function

c_src/unifex/unifex/logger.h
khamilowicz: That's right, removing

bundlex.exs
```diff
@@ -12,12 +12,24 @@ defmodule Unifex.BundlexProject do
       unifex: [
         deps: [shmex: :shmex],
         src_base: "unifex/nif/unifex",
-        sources: ["unifex.c", "payload.c"],
+        sources: ["unifex.c", "payload.c", "../../unifex/logger.c", "../../unifex/logger_nif.c"],
```
varsill, line 15: I don't think I understand why we need to use this `../../unifex/logger.c` 🤔

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+// like CNode - see the comment on send_log_func above for why one process-wide
+// value is sufficient). Optional: NULL if the backend supplies its own (e.g.
+// the NIF backend's fallback_env, see logger_nif.c).
+static void *global_env = NULL;
+
+static bool logger_active = false;
+
+void *unifex_logger_worker(void *arg);
+
+static void free_tags_copy(char **tags, unsigned int tags_length) {
```
varsill, line 47: why it's called `free_tags_copy` and not `free_tags`

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/time.h>
+#include <time.h>
+
+// Global queue instance
+static UnifexLoggerQueue queue;
+
+// Target process name
+static const char *target_pid_name = "Elixir.Unifex.Logger";
```
varsill, line 28: Let's call it `target_process_name`, PIDs don't have names I suppose :D

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+
+  pthread_mutex_destroy(&queue.mutex);
+  pthread_cond_destroy(&queue.cond);
+}
+
+bool unifex_log(UnifexLogLevel level, const char *message, const char **tags,
+                unsigned int tags_length) {
+  if (!message) {
+    return false;
+  }
```
varsill, line 129: Is it ever the case that `message` is `NULL`?

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+
+  pthread_mutex_lock(&queue.mutex);
+  bool full = queue.count >= UNIFEX_LOGGER_MAX_QUEUE_SIZE;
+  if (full) {
+    queue.dropped_count++;
+  }
```
varsill, line 135: I have an impression that we should put logic like this inside `queue` "module" functions instead of accessing all its internals in `logger.c`.
At the same time, since `queue` state needs to be synchronized with `logger`'s state (e.g. `logger_active` flag) I think it might be beneficial to move the `mutex` out of the `queue` to `logger` and treat the `queue` as a pure data structure (to be used with in a single-threaded environment)

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+      tags_alloc_failed = true;
+    } else {
+      for (unsigned int i = 0; i < tags_length; i++) {
+        tags_copy[i] = tags[i] ? strdup(tags[i]) : NULL;
+        if (tags[i] && !tags_copy[i]) {
+          tags_alloc_failed = true;
+        }
+      }
+    }
+  }
```
varsill, line 159: [NIT] I would move it to a helper function

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+
+  if (!message_copy || tags_alloc_failed) {
+    free_pending_message(message_copy, tags_copy, tags_length);
+    return false;
+  }
+
+  pthread_mutex_lock(&queue.mutex);
+
+  if (queue.count >= UNIFEX_LOGGER_MAX_QUEUE_SIZE) {
```
varsill, line 168: Haven't we just checked it in line :132?

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// Register the send function (to be called from backend initialization)
+void unifex_logger_register_send_func(UnifexLoggerSendFunc func) {
+  send_log_func = func;
+}
+
+void unifex_logger_set_env(void *env) { global_env = env; }
+
+void unifex_logger_init() {
```
varsill, line 89: If I understand correctly, `unifex_logger_init()`, which starts the logger thread, is called at NIF's load callback (i.e. when the native library is loaded).
Since the sources are statically linked into each library's NIF .so file, it means that we will have a separate logger thread for each library (and then e.g. the `dropped_count` is counted per library, and not globally), which will send messages to a single, "global" Elixir process responsible for logging. Wouldn't it be possible to have a single logger thread (e.g. shipped as a separate `libunifex_logger.so` that all the NIFs would link against)?

c_src/unifex/unifex/logger.c
```diff
@@ -0,0 +1,248 @@
...
+  pthread_cond_destroy(&queue.cond);
+}
+
+bool unifex_log(UnifexLogLevel level, const char *message, const char **tags,
+                unsigned int tags_length) {
+  if (!message) {
+    return false;
+  }
+
+  pthread_mutex_lock(&queue.mutex);
```
varsill, line 131: Shouldn't we check if the mutex still exists? (i.e. assert against `logger_active`).
Otherwise, I think there might be a RC at shutdown, when `unifex_logger_cleanup` was already called, but some other thread still calls `unifex_log`

varsill (changes requested):
This design with a separate logger thread is indeed quite complicated, but I think these are the benefits:
* ~we can log from resource destructors (Erlang docs state that e.g. enif_make_tuple cannot be called from a destructor~ [EDIT] since OTP-19 it's possible to allocate `enif` resources from resource destructor and use them with `enif_send` :D 
* logging doesn't require resource allocation on the "main" NIF thread, which might have an impact on time-sensitive NIFs
* we can easily unify logger against both NIFs and CNodes (with CNodes, sending to an Elixir process requires writing over a TCP socket which is blocking and I don't think we should be blocking in the "processing path")

We need to decide whether or not it outweights the cost of having the logger that complicated.
Also, I wonder if we could make sure that there is only a single logger thread per application (and not per library, like now) - see [this comment](https://github.com/membraneframework/unifex/pull/123/changes#r3757925252) for more info.

**[EDIT] After discussion with @FelonEkonom and @mat-hek we came to the conclusion, that this queue thread is overengineering - CNodes are no longer first-class citizen of Unifex and resource allocation on NIF takes some negligable time. Let's remove the logger thread at all and make `unifex_log` a simple `enif_send()`.**

